PR #144, opened 2022-12-27
various improvements for cross-platform development

mikeroberts3000:
This PR includes the following improvements to support cross-platform development, i.e., where the source code lives in a common folder that is accessed by multiple operating systems:
- changed `build_third_party_libs.py` to:
  - place all build artifacts in a platform-specific top-level folder, so build artifacts for each platform can co-exist and don't interfere with each other
  - generate a well-behaved include folder for Eigen (the current code exports the entire top-level libeigen GitHub folder as an include folder, so you could `#include <README.md>` in C++ code and it wouldn't be a compilation error, which is undesirable behavior)
  - use the generated include folder for Eigen when building RBDL (the current code assumes that Eigen is already exported in a standard location that is findable by the `FindEigen3.cmake` file that is included in the RBDL repository, which defeats the purpose of including Eigen among our `third_party` dependencies)
  - fixed deprecation warnings on Windows
  - fixed exception warnings on Windows
- changed `generate_config.py` to iterate over all the projects in `unreal_projects`, instead of requiring the user to specify a project on the command-line, thereby making it easier to use and more consistent with other build tools
- added code that improves our handling of symlinks, especially symlinks that are created on one platform and accessed on another
- cleaned up and refactored code

### DIFF
--- a/cpp/unreal_plugins/CoreUtils/Source/Config.cpp
+++ b/cpp/unreal_plugins/CoreUtils/Source/Config.cpp
@@ -24,10 +24,9 @@ void Config::initialize()
         // Read config file contents into a YAML::Node
         config_file = FPaths::ConvertRelativePathToFull(FPaths::ProjectDir().Append("Temp/config.yaml"));
         config_ = YAML::LoadFile(TCHAR_TO_UTF8(*config_file));
-    }
 
     // Otherwise assert
-    else {
+    } else {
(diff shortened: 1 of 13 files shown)


RachithP (approved):
Looks good!

Tested getting_started and generate_image_dataset examples on Windows platform and they work as expected.

PR #147, opened 2023-01-06
improvements to the OpenBotPawn interface and other cleanup tasks

mikeroberts3000:
At first glance, this PR looks more substantial than it actually is. This is a set of mostly cosmetic changes.

- I rearranged the order of the functions in `AOpenBotPawn` to be grouped more logically, changed the order of functions in the source file to match the header, and slightly modified the interface to be more clear and explicit.
- I introduced an `StdUtils::contains(...)` function to clean up usages of `std::find` throughout our code, which is overly verbose
- I introduced additional folder hierarchy in each plugin and project so our `#include` statements are more consistent and more logically organized
- I added `[SPEAR | filename]` at the beginning of all print statements so our debug output is more clearly delineated
- I removed the `SceneManager` plugin because we're not ready to use it. We can reintroduce it later when we're ready to use it, but for now there is no need for it to clutter up our code.


mikeroberts3000:
@RachithP

RachithP (approved):
Looks good.

Pending tasks:
- Sync to `main`.
- Remove SceneManager related content from `.gitignore`.

mikeroberts3000:
Good catch. Synced to `main` and removed `SceneManager` from `.gitignore`.

PR #148, opened 2023-01-12
tweaks to UrdfBot

mikeroberts3000:
- removed for loops to improve readability
 - fixed build file
 - added to gitignore


.gitignore
```diff
@@ -12,6 +12,10 @@ cpp/unreal_plugins/SimulationController/Binaries
 cpp/unreal_plugins/SimulationController/Intermediate
 cpp/unreal_plugins/SimulationController/ThirdParty
 
+cpp/unreal_plugins/UrdfBot/Binaries
+cpp/unreal_plugins/UrdfBot/Intermediate
+cpp/unreal_plugins/UrdfBot/ThirdParty
```
rwang15, line 17: `.gitignore` allows patterns such as `cpp/unreal_plugins/*/Binaries`. Is there any reason we dont take advantage of this feature?

rwang15 (approved):
minor question about gitignore. Reasonable changes in `UrdfParser`.

.gitignore
mikeroberts3000: There is not a good reason why we don't take advantage of this feature. This is a historical artifact due to my own ignorance when I was initially setting up the `.gitignore` file. I changed it according to your suggestion for each of our plugins. I didn't see any other opportunities to use wildcards, but I like that there is now an example usage in the `.gitignore` file because maybe there will be other opportunities in future.

PR #151, opened 2023-01-24
Python script to build pak files

RachithP:
The `pak` files generated by this script are not 100% functional. There is an issue with materials that is yet to be fixed.

In the meantime, I wanted to get feedback on the script.


RachithP:
Fixed the issue with materials not loading properly. Fix was to add the following to `DefaultGame.ini`
```
[/Script/UnrealEd.ProjectPackagingSettings]
bShareMaterialShaderCode=False
```

tools/build_pak_files.py
```diff
@@ -0,0 +1,100 @@
...
+    parser.add_argument("--unreal_project_dir", required=True)
+    parser.add_argument("--scene_id", required=True)
+    parser.add_argument("--platforms", nargs="*", required=True, help="Supported platform values are Windows, Linux, and Mac")
+    parser.add_argument("--output_dir", required=True)
+    args = parser.parse_args()
+
+    os.makedirs(args.output_dir, exist_ok=True)
+    
+    assert os.path.exists(args.unreal_binaries_dir)
+    unreal_editor_exe = os.path.realpath(os.path.join(args.unreal_binaries_dir, "Win64", "UE4Editor.exe"))
```
mikeroberts3000, line 25: This tool should be cross-platform. Why is building PAK files restricted to Windows?

tools/build_pak_files.py
```diff
@@ -0,0 +1,100 @@
...
+
+        # temporary text file used to generate the final pak file
+        pak_txt_path = os.path.join(args.output_dir, f"pak_data_{platform}.txt")
+
+        if os.path.exists(pak_txt_path):
+            os.remove(pak_txt_path)
+
+        for dir in content_dirs:
+            with open(pak_txt_path, mode='a') as fout:
+                for fpath in glob.glob(dir + "\**\*.*", recursive=True):
```
mikeroberts3000, line 75: use os.path.join instead of this style of concatenating

tools/build_pak_files.py
```diff
@@ -0,0 +1,100 @@
...
+import os
+import subprocess
+import sys
+
+
+if __name__ == '__main__':
+
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--unreal_binaries_dir", required=True)
+    parser.add_argument("--unreal_project_dir", required=True)
```
mikeroberts3000, line 16: unreal_project_dir should be optional and handled in the same way as in generate_config.py

tools/build_pak_files.py
```diff
@@ -0,0 +1,100 @@
...
+    parser.add_argument("--scene_id", required=True)
```
mikeroberts3000, line 17: scene_id should be optional

tools/build_pak_files.py
```diff
@@ -0,0 +1,100 @@
...
+    assert os.path.exists(args.unreal_project_dir)
+    _, project = os.path.split(args.unreal_project_dir)
+    uproject   = os.path.realpath(os.path.join(args.unreal_project_dir, project + ".uproject"))
+    
+    for platform in args.platforms:
+
+        cmd = [ 
+            unreal_editor_exe,
+            uproject,
+            "-run=Cook",
```
mikeroberts3000, line 37: Where do these command-line arguments come from? Are the possibilities documented anywhere?

tools/build_pak_files.py
```diff
@@ -0,0 +1,100 @@
...
+            "-unattended",
+            "-NoLogTimes",
+            "-UTF8Output"
+        ]
+        print(f"[SPEAR | build_pak_files.py] Executing: {' '.join(cmd)}")
+        cmd_result = subprocess.run(cmd, stdout=sys.stdout.buffer, stderr=sys.stdout.buffer)
+        assert cmd_result.returncode == 0
+
+        content_dirs = [
+            os.path.realpath(os.path.join(args.unreal_project_dir, "Saved", "Cooked", f"{platform}NoEditor", project,  "Content")),
```
mikeroberts3000, line 54: sort these dirs

tools/build_pak_files.py
```diff
@@ -0,0 +1,100 @@
...
+            os.path.realpath(os.path.join(args.unreal_project_dir, "Saved", "Cooked", f"{platform}NoEditor", "Engine", "Content", "BasicShapes")),
+            os.path.realpath(os.path.join(args.unreal_project_dir, "Saved", "Cooked", f"{platform}NoEditor", "Engine", "Content", "Functions", "Engine_MaterialFunctions02", "Texturing")),
+            os.path.realpath(os.path.join(args.unreal_project_dir, "Saved", "Cooked", f"{platform}NoEditor", "Engine", "Content", "Functions", "Engine_MaterialFunctions02", "Math")),
+            os.path.realpath(os.path.join(args.unreal_project_dir, "Saved", "Cooked", f"{platform}NoEditor", "Engine", "Content", "Functions", "MaterialLayerFunctions")),
+            os.path.realpath(os.path.join(args.unreal_project_dir, "Saved", "Cooked", f"{platform}NoEditor", "Engine", "Content", "Functions", "Engine_MaterialFunctions02", "Utility")),
+            os.path.realpath(os.path.join(args.unreal_project_dir, "Saved", "Cooked", f"{platform}NoEditor", "Engine", "Content", "EngineResources"))
+        ]
+
+        # construct output pak file path
+        output_pak = os.path.join(args.output_dir, f"Map_{args.scene_id}_{platform}.pak")
```
mikeroberts3000, line 65: pak_file

tools/build_pak_files.py
```diff
@@ -0,0 +1,100 @@
...
+
+        # temporary text file used to generate the final pak file
+        pak_txt_path = os.path.join(args.output_dir, f"pak_data_{platform}.txt")
```
mikeroberts3000, line 68: txt_file

tools/build_pak_files.py
```diff
@@ -0,0 +1,100 @@
...
+
+        if os.path.exists(pak_txt_path):
```
mikeroberts3000, line 70: don't remove, just write to a text file

tools/build_pak_files.py
mikeroberts3000, line 75: f instead of fpath

tools/build_pak_files.py
```diff
@@ -0,0 +1,100 @@
...
+            os.remove(pak_txt_path)
+
+        for dir in content_dirs:
```
mikeroberts3000, line 73: content_dir instead of dir

tools/build_pak_files.py
```diff
@@ -0,0 +1,100 @@
...
+            with open(pak_txt_path, mode='a') as fout:
```
mikeroberts3000, line 74: "a"

tools/build_pak_files.py
mikeroberts3000, line 75: use os.path.join instead of this style of concatenating

tools/build_pak_files.py
```diff
@@ -0,0 +1,100 @@
...
+                for fpath in glob.glob(dir + "\**\*.*", recursive=True):
+                    fpath = fpath.replace('\\', "/")
+                    mount_path = fpath.split(f"{platform}NoEditor")[1]
+                    fout.write(f'"{fpath}" "../../..{mount_path}" "" \n')
```
mikeroberts3000, line 78:
```
platform_dir = os.path.realpath(os.path.join(args.unreal_project_dir, "Saved", "Cooked", f"{platform}NoEditor"))
content_dirs = [
    os.path.join(platform_dir, ...),
    ... ]

...

assert content_file.startswith(platform_dir)

content_file = content_file.replace("\\", "/")
mount_file = posixpath.join("..", "..", "..", content_file.split(platform_dir)[1])
f.write(f'"{content_file}" "{mount_file}" "" \n'```
```

tools/build_pak_files.py
```diff
@@ -0,0 +1,100 @@
...
+        cmd_result = subprocess.run(cmd, stdout=sys.stdout.buffer, stderr=sys.stdout.buffer)
+        assert cmd_result.returncode == 0
+
+        if os.path.exists(output_pak):
+            print(f"[SPEAR | build_pak_files.py] Successfully built {output_pak}.")
+        else:
+            print(f"[SPEAR | build_pak_files.py] Could not build {output_pak} for {platform} platform.")
+
+        # remove the temporary text file
+        os.remove(pak_txt_path)
```
mikeroberts3000, line 100: don't remove

tools/build_pak_files.py
```diff
@@ -0,0 +1,100 @@
...
+            "-platform=" + platform,
+            "-multiprocess",
+            "-compress"
+        ]
+
+        print(f"[SPEAR | build_pak_files.py] Executing: {' '.join(cmd)}")
+        cmd_result = subprocess.run(cmd, stdout=sys.stdout.buffer, stderr=sys.stdout.buffer)
+        assert cmd_result.returncode == 0
+
+        if os.path.exists(output_pak):
```
mikeroberts3000, line 94: assert

mikeroberts3000:
- This script will build all content into exactly one pak file (not a pak file per scene), but the scene_id argument is used in a misleading way to name the pak file.
- This script also assumes that the project's Content folder has all the necessary data from Perforce, but this is undesirable. The script should create a symbolic link from the Perforce scene folder to the Content folder before building and remove it when finished.

tools/build_pak_files.py
```diff
@@ -0,0 +1,100 @@
...
+            "-unversioned",
+            "-stdout",
+            "-FullStdOutLogOutput",
+            "-CrashForUAT",
+            "-unattended",
+            "-NoLogTimes",
+            "-UTF8Output"
+        ]
+        print(f"[SPEAR | build_pak_files.py] Executing: {' '.join(cmd)}")
+        cmd_result = subprocess.run(cmd, stdout=sys.stdout.buffer, stderr=sys.stdout.buffer)
```
mikeroberts3000, line 50: I don't think setting stdout= and stderr= like this has any effect because you're just setting it to the default. Are you sure this has an effect?

tools/build_pak_files.py
RachithP: Donw

tools/build_pak_files.py
RachithP: Done

tools/build_pak_files.py
RachithP: Done

tools/build_pak_files.py
RachithP: Done

tools/build_pak_files.py
RachithP: Done

tools/build_pak_files.py
RachithP: Done

tools/build_pak_files.py
RachithP: Done

tools/build_pak_files.py
RachithP: Done

tools/build_pak_files.py
RachithP: Done

tools/build_pak_files.py
RachithP: Updated

tools/build_pak_files.py
RachithP: Done

tools/build_pak_files.py
RachithP: You are right, they did not have any effect. Removed them.

tools/build_pak_files.py
RachithP: Done

tools/build_pak_files.py
RachithP: Hmmm, currently our perforce project contains only 1 scene in content folder. As we add more scenes, the same content folder will grow and contain all the scenes. So, building a separate pak for each scene from this content folder would not be possible as far as I know (i.e., in this scenario, I'm not sure we can single out a scene for this pak generation process).

If building a separate pak for each scene is desirable (in the long run, with 300 or more scenes, I think this would be the better option), then we will have to change the current Perforce project directory structure, and update this script to generate pak file for each scene separately.

If building a single pak file containing all provided scenes is desirable, then, we do not need `scene_id` at all. We will name the final pak as something like `SpearSim-Content-v...-Win64.pak`.

What do you think?

RachithP:
> * This script will build all content into exactly one pak file (not a pak file per scene), but the scene_id argument is used in a misleading way to name the pak file.
> * This script also assumes that the project's Content folder has all the necessary data from Perforce, but this is undesirable. The script should create a symbolic link from each Perforce scene folder to the Content folder before building building a pak file if it doesn't already exist, and then remove the symlink when finished building if it needed to create one.

You are right. I've updated the script to handle multiple folders (each folder will contain contents for each scene) inside an outer folder `args.contents_dir`.

tools/build_pak_files.py
```diff
@@ -0,0 +1,129 @@
...
+        unreal_pak_bin    = os.path.realpath(os.path.join(args.unreal_binaries_dir, "Linux", "UnrealPak.sh"))
+    
+    if args.unreal_project_dir:
+        unreal_project_dirs = [os.path.realpath(args.unreal_project_dir)]
+    else:
+        unreal_projects_dir = os.path.realpath(os.path.join(os.path.dirname(os.path.realpath(__file__)), "..", "cpp", "unreal_projects"))
+        unreal_project_dirs = [ os.path.join(unreal_projects_dir, project) for project in os.listdir(unreal_projects_dir) ]
+    
+    assert os.path.exists(args.contents_dir)
+    content_dirs = [os.path.realpath(os.path.join(args.contents_dir, x)) for x in os.listdir(args.contents_dir)]
```
mikeroberts3000, line 42: Pad with whitespace for list comprehensions.

```
[ os.path.realpath(os.path.join(args.contents_dir, x)) for x in os.listdir(args.contents_dir) ]
```

tools/build_pak_files.py
```diff
@@ -0,0 +1,129 @@
...
+    for unreal_project_dir in unreal_project_dirs:
+
+        assert os.path.exists(unreal_project_dir)
+        _, project = os.path.split(unreal_project_dir)
+        uproject   = os.path.realpath(os.path.join(unreal_project_dir, project + ".uproject"))
+
+        for content_dir in content_dirs:
+
+            # remove existing Content dir
+            unreal_project_content_dir = os.path.join(unreal_project_dir, "Content")
```
mikeroberts3000, line 54: SpearSim already has a real physical `Content` dir (not a symlink) that is checked into GitHub. I think we need a different approach here, because I don't want this script to delete checked in files.

tools/build_pak_files.py
```diff
@@ -0,0 +1,129 @@
...
+
+                file_name = f"Map_{os.path.basename(content_dir)}_{platform}"
+
+                # construct path to the output pak file
+                pak_file = os.path.join(args.output_dir, file_name + ".pak")
+
+                # text file used to generate the final pak file
+                txt_path = os.path.join(args.output_dir, file_name + ".txt")
+
+                for i, content_dir in enumerate(content_dirs):
```
mikeroberts3000, line 106: This inner `content_dir` variable masks an existing `content_dir` variable. Is this intended?

tools/build_pak_files.py
```diff
@@ -0,0 +1,129 @@
...
+                    os.path.join(platform_dir, "Engine", "Content", "Functions", "Engine_MaterialFunctions02", "Texturing"),
+                    os.path.join(platform_dir, "Engine", "Content", "Functions", "Engine_MaterialFunctions02", "Utility"),
+                    os.path.join(platform_dir, "Engine", "Content", "Functions", "MaterialLayerFunctions"),
+                    os.path.join(platform_dir, project,  "Content")
+                ]
+
+                # create the output_dir
+                os.makedirs(args.output_dir, exist_ok=True)
+
+                file_name = f"Map_{os.path.basename(content_dir)}_{platform}"
```
mikeroberts3000, line 98: `pak_file_prefix`

tools/build_pak_files.py
```diff
@@ -0,0 +1,129 @@
...
+
+                # construct path to the output pak file
+                pak_file = os.path.join(args.output_dir, file_name + ".pak")
+
+                # text file used to generate the final pak file
+                txt_path = os.path.join(args.output_dir, file_name + ".txt")
```
mikeroberts3000, line 104: `txt_file` to match `pak_file`

tools/build_pak_files.py
```diff
@@ -0,0 +1,129 @@
...
+                    "-CrashForUAT",
+                    "-unattended",
+                    "-NoLogTimes",
+                    "-UTF8Output"
+                ]
+                print(f"[SPEAR | build_pak_files.py] Executing: {' '.join(cmd)}")
+                cmd_result = subprocess.run(cmd, check=True)
+
+                platform_dir = os.path.realpath(os.path.join(unreal_project_dir, "Saved", "Cooked", f"{platform}NoEditor"))
+                content_dirs = [
```
mikeroberts3000, line 84: This inner `content_dirs` variable masks an existing `content_dirs` variable. Is this intended?

tools/build_pak_files.py
```diff
@@ -0,0 +1,129 @@
...
+
+                # text file used to generate the final pak file
+                txt_path = os.path.join(args.output_dir, file_name + ".txt")
+
+                for i, content_dir in enumerate(content_dirs):
+                    with open(txt_path, mode="w" if i==0 else "a") as f:
+                        for content_file in glob.glob(os.path.join(content_dir, "**", "*.*"), recursive=True):
+                            assert content_file.startswith(platform_dir)
+                            content_file = content_file.replace('\\', "/")
+                            mount_file = posixpath.join("..", "..", f"..{content_file.split(platform + 'NoEditor')[1]}")
```
mikeroberts3000, line 111: Why are we setting `mount_file` to `../../..path/to/uasset` (note the missing `/` before `path`)?

tools/build_pak_files.py
```diff
@@ -0,0 +1,129 @@
...
+import spear
+import subprocess
+import sys
+
+
+if __name__ == '__main__':
+
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--contents_dir", required=True)
+    parser.add_argument("--unreal_binaries_dir", required=True)
```
mikeroberts3000, line 18: `unreal_engine_dir` to match `build_executable.py`

tools/build_pak_files.py
```diff
@@ -0,0 +1,129 @@
...
+    parser.add_argument("--platforms", nargs="*", required=True)
+    parser.add_argument("--output_dir", required=True)
+    parser.add_argument("--unreal_project_dir")
```
mikeroberts3000, line 21: It's fine with me if we hard-code `unreal_project_dir` to `SpearSim`, similar to how we do it in `build_executable.py`. I don't think it would ever happen that we'd want to build everything in `args.contents_dir` for every project in `unreal_projects`.

tools/build_pak_files.py
mikeroberts3000: Update: I've updated my belief here, I now think that it is ok to hard-code the unreal_project_dir to `SpearSim` because I think we'll always want to build specific content for specific unreal projects.

tools/build_pak_files.py
mikeroberts3000: I believe it is desirable to build a pak file for each scene, and we should have a Perforce folder for each scene. This will also be desirable when you want to edit a specific scene in the Unreal Editor. You should be able to create a symlink in `SpearSim/Content` that refers to a specific scene to facilitate editing that specific scene.

tools/build_pak_files.py
RachithP: I agree, updated code to handle only `SpearSim`.

tools/build_pak_files.py
RachithP: The output of `content_file.split(platform + 'NoEditor')[1]` has a trailing forward slash (`/`) such as `/SpearSim/Content/....`.

tools/build_pak_files.py
RachithP: Updated code to avoid this conflict.

tools/build_pak_files.py
RachithP: Done

tools/build_pak_files.py
RachithP: Done

tools/build_pak_files.py
RachithP: Updated code to avoid this conflict.

tools/build_pak_files.py
RachithP: Done

tools/build_pak_files.py
RachithP: I agree. Updated code to only handle `SpearSim`.

tools/build_pak_files.py
RachithP: The issue with `unreal_project_content_dir/scene_id -> perforce_content_dir/scene_id` is that this won't maintain the required dir structure of the `Content` dir. So, assets won't be able to find other linked assets, and cooking fails.

One way I could think of navigating this is to create symlinks for each dir inside `perforce_content/scene_id`. However, there is one issue here. The existing `Content` dir has a dir called `Materials` and all the `perforce_content/scene_id` dirs also contain a dir called `Materials`. This conflict in the name prevents us from creating a symlink to the `Materials` folder.

cpp/unreal_projects/SpearSim/Config/DefaultEngine.ini
```diff
@@ -27,8 +27,8 @@ r.EyeAdaptation.ExponentialTransitionDistance=0
 r.Streaming.FullyLoadUsedTextures=1
 
 [/Script/EngineSettings.GameMapsSettings]
-GameDefaultMap=/Game/DefaultMap.DefaultMap
-EditorStartupMap=/Game/DefaultMap.DefaultMap
+GameDefaultMap=/Game/Scenes/default/DefaultMap.DefaultMap
```
mikeroberts3000, line 30: `/Game/Scenes/default/Maps/DefaultMap.DefaultMap` to match the convention we established in our previous discussion.

python/spear/config/default_config.simulation_controller.yaml
```diff
@@ -5,8 +5,8 @@
 SIMULATION_CONTROLLER:
   IP: "127.0.0.1"
   PORT: 30000
-  WORLD_PATH_NAME: "" # full path to the map including the repeated name, e.g., "/Game/DefaultMap.DefaultMap"
-  LEVEL_NAME: ""      # full path to the map excluding the repeated name, e.g., "/Game/DefaultMap"
+  WORLD_PATH_NAME: "" # full path to the map including the repeated name, e.g., "/Game/Scenes/default/DefaultMap.DefaultMap"
```
mikeroberts3000, line 8: `/Game/Scenes/default/Maps/DefaultMap.DefaultMap`

tools/build_pak_files.py
```diff
@@ -0,0 +1,135 @@
...
+    scenes_dir = os.path.realpath(os.path.join(args.contents_dir, "Scenes"))
+
+    assert os.path.exists(shared_dir)
+    unreal_project_content_shared_dir = os.path.join(unreal_project_content_dir, "Shared")
+    if not spear.path_exists(unreal_project_content_shared_dir):
+        print(f"[SPEAR | build_pak_files.py] Creating symlink: {unreal_project_content_shared_dir} -> {shared_dir}")
+        os.symlink(shared_dir, unreal_project_content_shared_dir)
+
+    assert os.path.exists(scenes_dir)
+    scene_content_dirs = [ os.path.realpath(os.path.join(scenes_dir, x)) for x in os.listdir(scenes_dir) if x.startswith("kujiale") ]
```
mikeroberts3000, line 49: I don't think this script should know about specific "scene groups" (e.g., all the `kujiale_*` scenes). If anything, the user should be able to provide an optional scene name or a wild-card string that the scene IDs are matched against. There are examples of how to do this cleanly in the hypersim codebase, search for `fnmatch`.

tools/build_pak_files.py
```diff
@@ -0,0 +1,135 @@
...
+            print(f"[SPEAR | build_pak_files.py] File or directory or symlink exists, removing: {unreal_project_content_scene_dir}")
+            spear.remove_path(unreal_project_content_scene_dir)
+
+        # create symlink
+        print(f"[SPEAR | build_pak_files.py] Creating symlink: {unreal_project_content_scene_dir} -> {scene_content_dir}")
+        os.symlink(scene_content_dir, unreal_project_content_scene_dir)
+        
+        for platform in args.platforms:
+
+            cmd = [
```
mikeroberts3000, line 68: Where do all these arguments come from?

tools/build_pak_files.py
```diff
@@ -0,0 +1,135 @@
...
+                "-multiprocess",
+                "-compress"
+            ]
+            print(f"[SPEAR | build_pak_files.py] Executing: {' '.join(cmd)}")
+            subprocess.run(cmd, check=True)
+
+            assert os.path.exists(pak_file)
+            print(f"[SPEAR | build_pak_files.py] Successfully built {pak_file} for {platform} platform.")
+
+            print(f"[SPEAR | build_pak_files.py] Removing symlink {unreal_project_content_scene_dir}")
```
mikeroberts3000, line 132: `Removing symlink:` to match `Creating symlink:` above.

mikeroberts3000 (approved):
Some minor comments but otherwise looks great.

tools/build_pak_files.py
RachithP: Done

tools/build_pak_files.py
RachithP: Added a wild-card string based scene selection using `fnmatch`. Using `fnmatch` is pretty convenient and clean, thanks for the suggestion.

python/spear/config/default_config.simulation_controller.yaml
RachithP: Done

cpp/unreal_projects/SpearSim/Config/DefaultEngine.ini
RachithP: Done

tools/build_pak_files.py
RachithP: Updated to run on all supported platforms.

Upon testing on Linux, encountering an issue. I'm investigating it.

```bash
(spear) rachithp@rachithp:~/code/github/spear/tools (rachithp/build_paks)$ python build_pak_files.py --unreal_engine_dir /home/rachithp/code/github/unrealengine/UnrealEngine --contents_dir /media/rachithp/LaCie/spearsim_final_content_for_perforce --platforms Linux --output_dir /media/rachithp/LaCie/package_files_testing/linux --scene_names kujiale_0000
[SPEAR | build_pak_files.py] File or directory or symlink exists, removing: /home/rachithp/code/github/spear/cpp/unreal_projects/SpearSim/Content/Scenes/kujiale_0000
[SPEAR | build_pak_files.py] Creating symlink: /home/rachithp/code/github/spear/cpp/unreal_projects/SpearSim/Content/Scenes/kujiale_0000 -> /media/rachithp/LaCie/spearsim_final_content_for_perforce/Scenes/kujiale_0000
[SPEAR | build_pak_files.py] Executing: /home/rachithp/code/github/unrealengine/UnrealEngine/Engine/Binaries/Linux/UE4Editor /home/rachithp/code/github/spear/cpp/unreal_projects/SpearSim/SpearSim.uproject -run=Cook -TargetPlatform=LinuxNoEditor -fileopenlog -ddc=InstalledDerivedDataBackendGraph -unversioned -stdout -FullStdOutLogOutput -CrashForUAT -unattended -NoLogTimes -UTF8Output
/home/rachithp/code/github/unrealengine/UnrealEngine/Engine/Binaries/Linux/UE4Editor: symbol lookup error: /home/rachithp/code/github/unrealengine/UnrealEngine/Engine/Binaries/Linux/libUE4Editor-UnrealEd.so: undefined symbol: _ZN11FDDCCleanup8RunnableE
Traceback (most recent call last):
  File "/home/rachithp/code/github/spear/tools/build_pak_files.py", line 91, in <module>
    subprocess.run(cmd, check=True)
  File "/usr/lib/python3.9/subprocess.py", line 528, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command '['/home/rachithp/code/github/unrealengine/UnrealEngine/Engine/Binaries/Linux/UE4Editor', '/home/rachithp/code/github/spear/cpp/unreal_projects/SpearSim/SpearSim.uproject', '-run=Cook', '-TargetPlatform=LinuxNoEditor', '-fileopenlog', '-ddc=InstalledDerivedDataBackendGraph', '-unversioned', '-stdout', '-FullStdOutLogOutput', '-CrashForUAT', '-unattended', '-NoLogTimes', '-UTF8Output']' returned non-zero exit status 127.
```

tools/build_pak_files.py
RachithP: I'm not sure about all the params. I could only find some [here](https://docs.unrealengine.com/4.26/en-US/SharingAndReleasing/Deployment/Cooking/). @rwang15, do you have more information on this?

tools/build_pak_files.py
RachithP: Confirmed that it was an issue with the UE on my system. I reinstalled UE 4.26 and now I'm able to execute this script successfully. The Linux pak file generated from the script also works with generate_image_dataset example.

PR #152, opened 2023-01-25
Python script to automate codesign procedure for macOS executables

RachithP:
macOS executables built from Unreal Engine require some special steps to be ready for distribution. This script automates those steps.


tools/codesign_mac_executable.py
```diff
@@ -0,0 +1,232 @@
...
+        os.path.join(args.macos_executable, "Contents", "Resources", "RadioEffectUnit.component"),
+        os.path.join(args.macos_executable, "Contents", "UE4", "Engine", "Build", "Mac", "RadioEffectUnit")]
+
+    print(f"[SPEAR | codesign_mac_executable.py] Removing {radio_effect_units[0]}.")
+    shutil.rmtree(radio_effect_units[0], ignore_errors=True)
+
+    print(f"[SPEAR | codesign_mac_executable.py] Removing {radio_effect_units[1]}.")
+    shutil.rmtree(radio_effect_units[1], ignore_errors=True)
+
+    print(f"[SPEAR | codesign_mac_executable.py] Changing rpaths...")
```
mikeroberts3000, line 38: don't need the f since there you're not printing variables

tools/codesign_mac_executable.py
```diff
@@ -0,0 +1,232 @@
...
+import os
+import shutil
+import subprocess
+import time
+
+
+if __name__ == "__main__":
+
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--macos_executable", required=True)
```
mikeroberts3000, line 18: - rename to sign_executable.py
- rename to args.executable
- assert sys.platform == "darwin" at the beginning of the script

tools/codesign_mac_executable.py
```diff
@@ -0,0 +1,232 @@
...
+    shutil.rmtree(radio_effect_units[0], ignore_errors=True)
+
+    print(f"[SPEAR | codesign_mac_executable.py] Removing {radio_effect_units[1]}.")
+    shutil.rmtree(radio_effect_units[1], ignore_errors=True)
+
+    print(f"[SPEAR | codesign_mac_executable.py] Changing rpaths...")
+
+    # change current working directory to add relative rpaths 
+    cwd = os.getcwd()
+    os.chdir(os.path.join(args.macos_executable, ".."))
```
mikeroberts3000, line 42: - create a local variable for macos_executable = os.path.realpath(...)
- get realpath for os.path.join(args.macos_executable, "..")
- print(f"[SPEAR | ...] Changing working directory: {...}") it's worth printing that the script is doing this as a debugging tool, in case I need to execute the various cmd manually myself

tools/codesign_mac_executable.py
```diff
@@ -0,0 +1,232 @@
...
+        "-rpath",
+        "/Volumes/Work/Perforce/UE4/Engine/Binaries/ThirdParty/PhysX3/Mac",
+        os.path.join(executable_name, "Contents", "UE4", "Engine", "Binaries", "ThirdParty", "PhysX3", "Mac"),
+        os.path.join(executable_name, "Contents", "UE4", "Engine", "Binaries", "ThirdParty", "PhysX3", "Mac", "libApexFramework.dylib")]
+    print(f"[SPEAR | codesign_mac_executable.py] Executing: {' '.join(cmd)}")
+    cmd_result = subprocess.run(cmd)
+    assert cmd_result.returncode == 0
+
+    # revert current working directory
+    os.chdir(cwd)
```
mikeroberts3000, line 154: print that you're changing the working directory as above

tools/codesign_mac_executable.py
```diff
@@ -0,0 +1,232 @@
...
+
+    # files that need to be codesigned
+    files = [
```
mikeroberts3000, line 157: sign_files

tools/codesign_mac_executable.py
```diff
@@ -0,0 +1,232 @@
...
+        os.path.join(executable_name, "Contents", "UE4", "Engine", "Binaries", "ThirdParty", "PhysX3", "Mac"),
+        os.path.join(executable_name, "Contents", "UE4", "Engine", "Binaries", "ThirdParty", "PhysX3", "Mac", "libPxFoundation.dylib")]
+    print(f"[SPEAR | codesign_mac_executable.py] Executing: {' '.join(cmd)}")
+    cmd_result = subprocess.run(cmd)
+    assert cmd_result.returncode == 0
+
+    cmd = [
+        "install_name_tool",
+        "-rpath",
+        "/Volumes/Work/Perforce/UE4/Engine/Binaries/ThirdParty/PhysX3/Mac",
```
mikeroberts3000, line 76: Is this absolute path going to cause problems for me when I try to execute this script?

tools/codesign_mac_executable.py
```diff
@@ -0,0 +1,232 @@
...
+    # change current working directory to add relative rpaths 
+    cwd = os.getcwd()
+    os.chdir(os.path.join(args.macos_executable, ".."))
+
+    executable_name = os.path.basename(args.macos_executable)
+
+    cmd = [
+        "install_name_tool",
+        "-add_rpath",
+        os.path.join(executable_name, "Contents", "UE4", "Engine", "Binaries", "ThirdParty", "Ogg", "Mac"),
```
mikeroberts3000, line 49: try to put these paths in a list or a dict so you can do this in a loop instead of repeating everything explicitly

tools/codesign_mac_executable.py
```diff
@@ -0,0 +1,232 @@
...
+        os.path.join(args.macos_executable, "Contents", "UE4", "Engine", "Binaries", "ThirdParty", "PhysX3", "Mac", "libPxFoundation.dylib"),
+        os.path.join(args.macos_executable, "Contents", "UE4", "Engine", "Binaries", "ThirdParty", "PhysX3", "Mac", "libAPEX_Clothing.dylib"),
+        os.path.join(args.macos_executable, "Contents", "UE4", "Engine", "Binaries", "ThirdParty", "PhysX3", "Mac", "libPxPvdSDK.dylib"),
+        os.path.join(args.macos_executable, "Contents", "UE4", "Engine", "Binaries", "ThirdParty", "PhysX3", "Mac", "libNvCloth.dylib"),
+        os.path.join(args.macos_executable, "Contents", "UE4", "Engine", "Binaries", "ThirdParty", "PhysX3", "Mac", "libPhysX3Cooking.dylib"),
+        os.path.join(args.macos_executable, "Contents", "UE4", "Engine", "Binaries", "ThirdParty", "PhysX3", "Mac", "libAPEX_Legacy.dylib"),
+        os.path.join(args.macos_executable, "Contents", "UE4", "Engine", "Binaries", "ThirdParty", "PhysX3", "Mac", "libApexFramework.dylib"),
+        os.path.join(args.macos_executable, "Contents", "MacOS", os.path.splitext(executable_name)[0])]
```
mikeroberts3000, line 169: we refer to Contents/MacOS differently in env.py, let's do the same thing here for consistency

tools/codesign_mac_executable.py
```diff
@@ -0,0 +1,232 @@
...
+    for file in files:
+        cmd = ["sudo", "codesign", "-f", "-s", "-v", "--options", "runtime", "--timestamp", "--entitlements", args.entitlements_file, "--sign", f"Developer ID Application: {args.developer_id}", file]
+        print(f"[SPEAR | codesign_mac_executable.py] Executing: {' '.join(cmd)}")
+        cmd_result = subprocess.run(cmd)
+        assert cmd_result.returncode == 0
+
+    os.makedirs(args.output_dir, exist_ok=True)
+
+    # create a zip file for notarization
+    temp_zip = os.path.join(args.output_dir, f"{os.path.splitext(os.path.basename(args.macos_executable))[0]}-temp.zip")
```
mikeroberts3000, line 182: write this to a user-specifiable temp folder that defaults to tmp

tools/codesign_mac_executable.py
```diff
@@ -0,0 +1,232 @@
...
+    assert cmd_result.returncode == 0
+
+    # send the zip file for notarization 
+    cmd = ["xcrun", "altool", "--notarize-app", "--primary-bundle-id", "org.embodiedaifoundation.spear", "--username", args.apple_username, "--password", args.apple_password, "--file", temp_zip]
+    print(f"[SPEAR | codesign_mac_executable.py] Executing: {' '.join(cmd)}")
+    ps = subprocess.Popen(cmd, stdout=subprocess.PIPE)
+    output = subprocess.check_output(["grep", "RequestUUID = "], stdin=ps.stdout, text=True)
+    ps.wait()
+    ps.stdout.close()
+    request_uuid = " ".join(output[14:].split())
```
mikeroberts3000, line 195: why not split on "RequestUUID = ", similar to how you do in the build pak file script, instead of doing this brittle and not-self-documenting indexing trick [14:]

tools/codesign_mac_executable.py
```diff
@@ -0,0 +1,232 @@
...
+        os.path.join(args.macos_executable, "Contents", "UE4", "Engine", "Binaries", "ThirdParty", "PhysX3", "Mac", "libNvCloth.dylib"),
+        os.path.join(args.macos_executable, "Contents", "UE4", "Engine", "Binaries", "ThirdParty", "PhysX3", "Mac", "libPhysX3Cooking.dylib"),
+        os.path.join(args.macos_executable, "Contents", "UE4", "Engine", "Binaries", "ThirdParty", "PhysX3", "Mac", "libAPEX_Legacy.dylib"),
+        os.path.join(args.macos_executable, "Contents", "UE4", "Engine", "Binaries", "ThirdParty", "PhysX3", "Mac", "libApexFramework.dylib"),
+        os.path.join(args.macos_executable, "Contents", "MacOS", os.path.splitext(executable_name)[0])]
+
+    assert os.path.exists(args.entitlements_file)
+
+    for file in files:
+        cmd = ["sudo", "codesign", "-f", "-s", "-v", "--options", "runtime", "--timestamp", "--entitlements", args.entitlements_file, "--sign", f"Developer ID Application: {args.developer_id}", file]
```
mikeroberts3000, line 174: break these long cmd up on multiple lines

tools/codesign_mac_executable.py
```diff
@@ -0,0 +1,232 @@
...
+    print(f"[SPEAR | codesign_mac_executable.py] Executing: {' '.join(cmd)}")
+    ps = subprocess.Popen(cmd, stdout=subprocess.PIPE)
+    output = subprocess.check_output(["grep", "RequestUUID = "], stdin=ps.stdout, text=True)
+    ps.wait()
+    ps.stdout.close()
+    request_uuid = " ".join(output[14:].split())
+    print(f"[SPEAR | codesign_mac_executable.py] Request UUID: {request_uuid}")
+
+    # check the status of notarization request
+    cmd = [
```
mikeroberts3000, line 199: no need to break this up on multiple lines

tools/codesign_mac_executable.py
```diff
@@ -0,0 +1,232 @@
...
+        "xcrun", "altool", "--notarization-info", request_uuid, "--username", args.apple_username, "--password", args.apple_password]
+    print(f"[SPEAR | codesign_mac_executable.py] Executing: {' '.join(cmd)}")
+    output = "             in progress"
```
mikeroberts3000, line 202: - what is going on here? why are you setting `output = "             in progress"`
- again, I'd like to see the full output of this program as it executes, which I'm not sure is happening here

tools/codesign_mac_executable.py
```diff
@@ -0,0 +1,232 @@
...
+    start_time = time.time()
+    elapsed_time = time.time() - start_time
+    while " ".join(output[13:].split()) == "in progress" and elapsed_time<args.wait_time_seconds:
```
mikeroberts3000, line 205: - this loop is confusing and not self-documenting
- `elapsed_time < args.wait_time_seconds`

tools/codesign_mac_executable.py
mikeroberts3000, line 49: what happens when you run this part of the script repeatedly? are you adding redundant rpaths to the executable or these dylibs?

tools/codesign_mac_executable.py
```diff
@@ -0,0 +1,232 @@
...
+        ps = subprocess.Popen(cmd, stdout=subprocess.PIPE)
+        output = subprocess.check_output(["grep", "     Status:"], stdin=ps.stdout, text=True)
+        ps.wait()
+        ps.stdout.close()
+        time.sleep(20)
+        elapsed_time = time.time() - start_time
+        print(f"[SPEAR | codesign_mac_executable.py] Waiting to get more information on the notarization request...")
+    
+    assert elapsed_time <= args.wait_time_seconds, f"[SPEAR | codesign_mac_executable.py] ERROR: Exceeded provided wait time ({args.wait_time_seconds}s) for notarization. Please complete this procedure after notarization is complete."
```
mikeroberts3000, line 214: - print this error, then assert false
- since this error might happen, you should provide an option to run the end of this script, assuming the user has attempted it before and it timed out (e.g., don't try to hack the rpaths and upload the executable, just skip to this step if the user specifies a uuid)

tools/codesign_mac_executable.py
```diff
@@ -0,0 +1,232 @@
...
+    
+    # staple the executable
+    cmd = ["xcrun", "stapler", "staple", args.macos_executable]
+    print(f"[SPEAR | codesign_mac_executable.py] Executing: {' '.join(cmd)}")
+    cmd_result = subprocess.run(cmd)
+    assert cmd_result.returncode == 0
+
+    print(f"[SPEAR | codesign_mac_executable.py] Removing the temporary zip file {temp_zip}")
```
mikeroberts3000, line 222: don't remove, just write to a user-specifiable temp folder

tools/codesign_mac_executable.py
```diff
@@ -0,0 +1,232 @@
...
+    # staple the executable
+    cmd = ["xcrun", "stapler", "staple", args.macos_executable]
+    print(f"[SPEAR | codesign_mac_executable.py] Executing: {' '.join(cmd)}")
+    cmd_result = subprocess.run(cmd)
+    assert cmd_result.returncode == 0
+
+    print(f"[SPEAR | codesign_mac_executable.py] Removing the temporary zip file {temp_zip}") 
+    os.remove(temp_zip)
+
+    # zip the stapled executable for distribution
```
mikeroberts3000, line 225: I think zipping the final executable for distribution should happen in a separate script, because this happens in a separate script for the other platforms.

tools/codesign_mac_executable.py
mikeroberts3000, line 49: Have you tested this on multiple mac machines? I think this may actually prevent the executable from running on other machines because these rpaths will only be valid for the machine where you're executing the script.

tools/codesign_mac_executable.py
```diff
@@ -0,0 +1,232 @@
...
+    parser.add_argument("--entitlements_file", required=True)
+    parser.add_argument("--developer_id", required=True)
+    parser.add_argument("--output_dir", required=True)
+    parser.add_argument("--apple_username", required=True)
+    parser.add_argument("--apple_password", required=True)
+    parser.add_argument("--wait_time_seconds", type=float, default=600.0)
+    args = parser.parse_args()
+
+    assert os.path.exists(args.macos_executable)
+    radio_effect_units = [
```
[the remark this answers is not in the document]
RachithP: Thought of grouping similar folders into an array. But, in this case, they are better being separate.

tools/codesign_mac_executable.py
RachithP: Done

tools/codesign_mac_executable.py
RachithP: Done

tools/codesign_mac_executable.py
RachithP: Done

tools/codesign_mac_executable.py
RachithP: Done

tools/codesign_mac_executable.py
RachithP: Done

tools/codesign_mac_executable.py
RachithP: We are removing this absolute path from the dylibs.

tools/codesign_mac_executable.py
RachithP: Done

tools/codesign_mac_executable.py
RachithP: I like this better, and maybe we should change the way we are handling this in env.py, because in env.py we are unnecessarily complicating the code to get the path to the executable.
What do you think?

tools/codesign_mac_executable.py
RachithP: Agreed, changed to `request_uuid = output.split("RequestUUID = ")[1].strip()` which is more clearer.

RachithP: Also, printing the piped output for more clarity.

tools/codesign_mac_executable.py
RachithP: Done

tools/codesign_mac_executable.py
RachithP: Done

tools/codesign_mac_executable.py
RachithP: It would through an error. Something like this

```bash
[SPEAR | sign_executable.py] Changing rpaths...
[SPEAR | sign_executable.py] Chaning working directory to /Users/rachithp/code/spear-mac-build-system/sign-app/SpearSim-Mac-Shipping.
[SPEAR | sign_executable.py] Executing: install_name_tool -add_rpath SpearSim-Mac-Shipping.app/Contents/UE4/Engine/Binaries/ThirdParty/Ogg/Mac SpearSim-Mac-Shipping.app/Contents/UE4/Engine/Binaries/ThirdParty/Ogg/Mac/libogg.dylib
error: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/install_name_tool: for: SpearSim-Mac-Shipping.app/Contents/UE4/Engine/Binaries/ThirdParty/Ogg/Mac/libogg.dylib (for architecture x86_64) option "-add_rpath SpearSim-Mac-Shipping.app/Contents/UE4/Engine/Binaries/ThirdParty/Ogg/Mac" would duplicate path, file already has LC_RPATH for: SpearSim-Mac-Shipping.app/Contents/UE4/Engine/Binaries/ThirdParty/Ogg/Mac
error: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/install_name_tool: for: SpearSim-Mac-Shipping.app/Contents/UE4/Engine/Binaries/ThirdParty/Ogg/Mac/libogg.dylib (for architecture arm64) option "-add_rpath SpearSim-Mac-Shipping.app/Contents/UE4/Engine/Binaries/ThirdParty/Ogg/Mac" would duplicate path, file already has LC_RPATH for: SpearSim-Mac-Shipping.app/Contents/UE4/Engine/Binaries/ThirdParty/Ogg/Mac
Traceback (most recent call last):
  File "sign_executable.py", line 68, in <module>
    assert cmd_result.returncode == 0
AssertionError
```

tools/codesign_mac_executable.py
RachithP: Agreed, changing this.

tools/codesign_mac_executable.py
RachithP: Sloppy implementation, changing this similar to `RequestUUID` above.

tools/codesign_mac_executable.py
RachithP: Done

tools/codesign_mac_executable.py
RachithP: I think we can do this when we integrate the two scripts, for e.g., after the executable is built using the build script, call the sign_executable.py on macOS binary. For now, maybe it will be good to have a final zip file as a product of the script.

tools/codesign_mac_executable.py
RachithP: As an aside, I tested out something similar, instead of using `executable_name`, I used `args.executable` i.e. an absolute path and placed the binary in a different location from that of `args.executable` and it did not work on the same machine itself. This is why I'm confident that the relative path works.

tools/codesign_mac_executable.py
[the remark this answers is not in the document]
RachithP: I added it to the folder.

tools/codesign_mac_executable.py
RachithP: I tried to read `stdout` (something like 
```python 
ps = subprocess.Popen(cmd, stdout=subprocess.PIPE)
for line in ps.stdout:
    print(f"[SPEAR | sign_executable.py] {line.decode()}", end="")
output = subprocess.check_output(["grep", "     Status:"], stdin=ps.stdout, text=True)
```
, but this causes issues with the subsequent `check_output` cmd. If I try to read like this
```python 
ps = subprocess.Popen(cmd, stdout=subprocess.PIPE)
output = subprocess.check_output(["grep", "     Status:"], stdin=ps.stdout, text=True)
for line in ps.stdout:
    print(f"[SPEAR | sign_executable.py] {line.decode()}", end="")
```
 I'm not able to, i.e. no output. 

Not sure how to go about this. Any ideas?

tools/codesign_macOS_executable/sign_executable.py
```diff
@@ -0,0 +1,177 @@
...
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--executable", required=True)
+    parser.add_argument("--developer_id", required=True)
+    parser.add_argument("--output_dir", required=True)
+    parser.add_argument("--tmp_dir", required=True)
+    parser.add_argument("--apple_username", required=True)
+    parser.add_argument("--apple_password", required=True)
+    parser.add_argument("--entitlements_file", default=os.path.join(os.path.dirname(os.path.realpath(__file__)), "entitlements.plist"))
+    parser.add_argument("--wait_time_seconds", type=float, default=600.0)
+    parser.add_argument("--request_uuid")
```
mikeroberts3000, line 29: - make it so the default behavior is to do this
- maybe make the parameter called "skip_request_uuid" and use action="store_true" to make it a bool

tools/codesign_macOS_executable/sign_executable.py
```diff
@@ -0,0 +1,177 @@
...
+    start_time = time.time()
+    elapsed_time = time.time() - start_time
+    while output == "in progress" and elapsed_time < args.wait_time_seconds:
+        ps = subprocess.Popen(cmd, stdout=subprocess.PIPE)
+        output = subprocess.check_output(["grep", "     Status:"], stdin=ps.stdout, text=True)
+        output = output.split("     Status:")[1].strip()
+        ps.wait()
+        ps.stdout.close()
+        elapsed_time = time.time() - start_time
+        print(f"[SPEAR | sign_executable.py] Waiting to get more information on the notarization request uuid: {request_uuid}, current status is {output} ...")
```
mikeroberts3000, line 158: - be consistent with these print statements
- Waiting to get more information on the notarization request UUID {request_uuid}, current status is {output}...
- Exceeded maximum wait time ({args.wait_time_seconds}s) for request UUID {request_uuid}, please complete the rest of the procedure after notarization is complete."

tools/codesign_macOS_executable/sign_executable.py
```diff
@@ -0,0 +1,177 @@
...
+                args.entitlements_file, "--sign", f"Developer ID Application: {args.developer_id}", file]
+            print(f"[SPEAR | sign_executable.py] Executing: {' '.join(cmd)}")
+            cmd_result = subprocess.run(cmd)
+            assert cmd_result.returncode == 0
+
+        os.makedirs(args.output_dir, exist_ok=True)
+        os.makedirs(args.tmp_dir, exist_ok=True)
+
+        # create a zip file for notarization
+        notrn_zip = os.path.join(args.tmp_dir, f"{os.path.splitext(os.path.basename(executable))[0]}.zip")
```
mikeroberts3000, line 127: notarization_zip

mikeroberts3000: You already assign os.path.basename(executable) to another variable, so use the variable you created of repeating os.path.basename(executable)

tools/codesign_mac_executable.py
mikeroberts3000: Ok sure, then let's change env.py. Here, you're assuming that the "internal" executable name is the same as the app name, whereas env.py doesn't make that assumption. But the way you're doing it here is more straightforward than env.py.

tools/codesign_mac_executable.py
mikeroberts3000: https://stackoverflow.com/questions/13398261/how-to-run-a-subprocess-with-python-wait-for-it-to-exit-and-get-the-full-stdout

tools/codesign_mac_executable.py
mikeroberts3000: Makes sense.

tools/codesign_mac_executable.py
mikeroberts3000: I'm not sure exactly how to do it, but it should definitely be possible to get the full output and the return code.

tools/codesign_mac_executable.py
mikeroberts3000: call the folder sign_executable

tools/codesign_mac_executable.py
mikeroberts3000: In this case, copy the executable to the temp folder and work in the temp folder, so you can assume you're working with a pristine executable instead of modifying it in-place.

tools/codesign_mac_executable.py
mikeroberts3000: I disagree. It's better for this script to have a more narrow responsibility. It should just "normalize" the mac executable so it can be processed by our other scripts in exactly the same way as the other platforms.

tools/codesign_mac_executable.py
mikeroberts3000: Ok that makes sense. I think I misunderstood the previous code.

tools/codesign_mac_executable.py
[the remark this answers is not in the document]
mikeroberts3000: Ok.

tools/codesign_macOS_executable/sign_executable.py
RachithP: Done

tools/codesign_macOS_executable/sign_executable.py
RachithP: Done

tools/codesign_macOS_executable/sign_executable.py
RachithP: Done

tools/codesign_mac_executable.py
RachithP: okay, removed.

tools/codesign_mac_executable.py
RachithP: Done

tools/codesign_mac_executable.py
RachithP: Good point, changed this.

tools/codesign_mac_executable.py
RachithP: Updated env.py

tools/codesign_mac_executable.py
RachithP: Done

tools/sign_executable/sign_executable.py
```diff
@@ -0,0 +1,178 @@
...
+
+
+if __name__ == "__main__":
+
+    assert sys.platform == "darwin"
+
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--input_dir", required=True)
+    parser.add_argument("--output_dir", required=True)
+    parser.add_argument("--temp_dir", required=True)
```
mikeroberts3000, line 24: default to "tmp" to match your other PR

tools/codesign_mac_executable.py
RachithP: I was able to solve this. Updated the code.

tools/sign_executable/sign_executable.py
```diff
@@ -0,0 +1,178 @@
...
+        print(f"[SPEAR | sign_executable.py] Exceeded maximum wait time ({args.wait_time_seconds}s) for request UUID {request_uuid}. Please complete the rest of the procedure after notarization is complete.")
+        assert False
+            
+    # staple the executable
+    cmd = ["xcrun", "stapler", "staple", executable]
+    print(f"[SPEAR | sign_executable.py] Executing: {' '.join(cmd)}")
+    cmd_result = subprocess.run(cmd)
+    assert cmd_result.returncode == 0
+
+    print(f"[SPEAR | sign_executable.py] Done: {executable} is successfully codesigned!")
```
mikeroberts3000, line 184: print(f"[SPEAR | sign_executable.py] {executable} has been successfully signed.")
printf("[SPEAR | sign_executable.py] Done.")

tools/sign_executable/sign_executable.py
```diff
@@ -0,0 +1,178 @@
...
+                dylib]
+            print(f"[SPEAR | sign_executable.py] Executing: {' '.join(cmd)}")
+            cmd_result = subprocess.run(cmd)
+            assert cmd_result.returncode == 0
+
+        # revert current working directory
+        print(f"[SPEAR | sign_executable.py] Changing working directory to {cwd}.")
+        os.chdir(cwd)
+
+        # files that need to be codesigned
```
mikeroberts3000, line 107: code-signed

mikeroberts3000 (approved):
Looks great.

tools/sign_executable/sign_executable.py
RachithP: Updated

tools/sign_executable/sign_executable.py
RachithP: Updated

tools/sign_executable/sign_executable.py
RachithP: Updated

PR #154, opened 2023-02-02
Python script automating build process

RachithP:
This script automates the build process, starting from cloning this repo, setting up required python envs and packages, to building the `SpearSim` executable and packing it into a zip file.


tools/build.py
```diff
@@ -0,0 +1,150 @@
...
+    parser.add_argument("--conda_path")
+    parser.add_argument("--config_mode", default="Shipping")
+    parser.add_argument("--unreal_engine_dir", required=True)
+    parser.add_argument("--num_parallel_jobs", type=int, default=1)
+    parser.add_argument("--target_platform", required=True, help="Values accepted are Win64, Linux, Mac")
+    parser.add_argument("--output_dir", required=True)
+    parser.add_argument("--tag", required=True)
+    args = parser.parse_args()
+
+    assert args.target_platform in ["Linux", "Mac"] and args.conda_path
```
mikeroberts3000, line 24: Don't want to maintain these help strings for all of our scripts, so let's remove. And they're hard to keep up-to-date, e.g., why does the assert not match the help string?

tools/build.py
```diff
@@ -0,0 +1,150 @@
...
+import argparse
+import os
+import shutil
+import subprocess
+import sys
+
+if __name__ == "__main__":
+
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--clone_dir", required=True)
```
mikeroberts3000, line 14: temp_dir

tools/build.py
```diff
@@ -0,0 +1,150 @@
...
+        assert False
+    print(f"[SPEAR | build.py] Executing: {' '.join(cmd)}")
+    cmd_result = subprocess.run(cmd, shell=True)
+    assert cmd_result.returncode == 0
+
+    # install msgpack-rpc-python package in the newly created python env
+    if sys.platform == "win32":
+        cmd = ["conda", "activate", "spear-env&", "pip", "install", "-e", "third_party/msgpack-rpc-python"]
+    elif sys.platform in ["linux", "darwin"]:    
+        cmd = [f". {args.conda_path}; conda activate spear-env; pip install -e third_party/msgpack-rpc-python"]
```
mikeroberts3000, line 54: It's fine to assume that "conda" is available on the command-line, why do we need conda_path here but not above?

tools/build.py
```diff
@@ -0,0 +1,150 @@
...
+    assert cmd_result.returncode == 0
+
+    os.makedirs(args.output_dir, exist_ok=True)
+
+    # change working directory to the cloned directory above
+    cwd = os.getcwd()
+    print(f"[SPEAR | build.py] Changing current working dir from {cwd} to {args.clone_dir}")
+    os.chdir(args.clone_dir)
+
+    # create a new virtual python env called spear-env
```
mikeroberts3000, line 39: - call this script package_executable.py
- don't call the environment spear-env because that is what I usually call it on my machine. Call it spear-package-executable-tmp-env

tools/build.py
```diff
@@ -0,0 +1,150 @@
...
+        "-project=" + os.path.join(args.clone_dir, "cpp", "unreal_projects", "SpearSim", "SpearSim.uproject"), 
+        "-build",
+        "-cook",
+        "-stage",
+        "-package",
+        "-archive",
+        "-pak",
+        "-targetplatform=" + args.target_platform,
+        "-target=SpearSim",
+        "-archivedirectory=" + args.output_dir,
```
mikeroberts3000, line 126: The archive directory should always be in the temp folder, with -Shipping or -Development

tools/build.py
```diff
@@ -0,0 +1,150 @@
...
+    cmd_result = subprocess.run(cmd)
+    assert cmd_result.returncode == 0
+
+    # change working directory
+    print(f"[SPEAR | build.py] Changing current working dir from {os.getcwd()} to {cwd}")
+    os.chdir(cwd)
+
+    print(f"[SPEAR | build.py] Successfully built SpearSim for {args.target_platform} platform with {args.config_mode} mode.")
+
+    # zip the exectuable for distribution
```
mikeroberts3000, line 139: need to code sign on mac

tools/build.py
```diff
@@ -0,0 +1,150 @@
...
+if __name__ == "__main__":
+
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--clone_dir", required=True)
+    parser.add_argument("--conda_path")
+    parser.add_argument("--config_mode", default="Shipping")
+    parser.add_argument("--unreal_engine_dir", required=True)
+    parser.add_argument("--num_parallel_jobs", type=int, default=1)
+    parser.add_argument("--target_platform", required=True, help="Values accepted are Win64, Linux, Mac")
+    parser.add_argument("--output_dir", required=True)
```
mikeroberts3000, line 20: output_dir can default to "."

tools/build.py
```diff
@@ -0,0 +1,150 @@
...
+
+    # change working directory
+    print(f"[SPEAR | build.py] Changing current working dir from {os.getcwd()} to {cwd}")
+    os.chdir(cwd)
+
+    print(f"[SPEAR | build.py] Successfully built SpearSim for {args.target_platform} platform with {args.config_mode} mode.")
+
+    # zip the exectuable for distribution
+    if sys.platform == "win32":
+        file = shutil.make_archive(base_name=os.path.join(args.output_dir, f"SpearSim-{args.tag}-{args.target_platform}-{args.config_mode}"), format='zip', root_dir=os.path.join(args.output_dir, "WindowsNoEditor"))
```
mikeroberts3000, line 141: can move make_archive out of the if statement

tools/build.py
```diff
@@ -0,0 +1,150 @@
...
+        cmd = ["python", "build_third_party_libs.py", "--num_parallel_jobs", f"{args.num_parallel_jobs}"]
+    elif sys.platform in ["linux", "darwin"]:    
+        cmd = [f"python build_third_party_libs.py --num_parallel_jobs {args.num_parallel_jobs}"]
+    else:
+        assert False
+    print(f"[SPEAR | build.py] Executing: {' '.join(cmd)}")
+    cmd_result = subprocess.run(cmd, shell=True)
+    assert cmd_result.returncode == 0
+
+    if sys.platform == "win32":
```
mikeroberts3000, line 86: I think it would be better for this script to do less setup. It should just: clone the repo, create symbolic links, generate a config, build, codesign (mac only), zip. For now, it should assume that everything else is already set up.

tools/build.py
RachithP: I agree that help strings are hard to keep up-to date. But, for the argument `--target_platform`, we need to provide from a fix set of strings such as 'Win64', 'Linux', 'Mac' as this is being passed to and interpreted by Unreal's build cmd. These values need not change so it might be helpful to have the help string in this case.

Another way would be to assign these strings (Win64, Linux, Mac) to a local variable as part of the code, based on `--target_platform`. I think help string way is better than this.

I did modify the assert statement.

tools/build.py
RachithP: done

tools/build.py
RachithP: When we use `shell=True` on unix systems, it defaults to `/bin/sh`. If your conda setup is not configured to `/bin/sh`'s profile (e.g., on mac, default is `/bin/zsh`, so your conda setup most likely will be configured for `zsh`'s profile), you won't be able to use conda commands. So sourcing conda.sh makes it so that conda can be used on any shell.

The other option would be take as input an argument (args.shell) to the shell's path such as `/bin/sh` or `/bin/zsh` so that we don't need conda_path or the need to source conda.sh as well.

I prefer the second option, what do you think?

tools/build.py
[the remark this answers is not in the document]
RachithP: removed the steps to create a conda env

RachithP: Agreed, changed.

tools/build.py
RachithP: Done

tools/build.py
RachithP: Added it as todo, I'll integrate it in this PR once https://github.com/isl-org/spear/pull/152 is merged.

tools/build.py
RachithP: Agreed.

tools/build.py
RachithP: Done.

tools/build.py
RachithP: Agreed, updated the script accordingly.

tools/build.py
mikeroberts3000: You could make the same argument for many of our command-line arguments. Let's remove the help string.

tools/build.py
mikeroberts3000: I thought we agreed to assume that your conda env is already set up.

tools/build.py
mikeroberts3000: I don't think you need to invoke conda before calling subprocess.run.

tools/package_executable.py
```diff
@@ -0,0 +1,129 @@
...
+import os
+import shutil
+import subprocess
+import sys
+
+if __name__ == "__main__":
+
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--temp_dir", required=True)
+    parser.add_argument("--conda_env", required=True)
```
mikeroberts3000, line 15: No need for this conda stuff

tools/package_executable.py
```diff
@@ -0,0 +1,129 @@
...
+    parser.add_argument("--unreal_engine_dir", required=True)
+    parser.add_argument("--num_parallel_jobs", type=int, default=1)
+    parser.add_argument("--target_platform", required=True, help="Values accepted are Win64, Linux, Mac")
+    parser.add_argument("--output_dir", default=os.path.dirname(os.path.realpath(__file__)))
+    parser.add_argument("--output_tag", required=True)
+    args = parser.parse_args()
+
+    repo_dir = os.path.join(os.path.realpath(args.temp_dir), "spear")
+
+    # need conda_path on unix systems
```
mikeroberts3000, line 27: No need for this

tools/package_executable.py
```diff
@@ -0,0 +1,129 @@
...
+    if sys.platform in [ "darwin", "linux"]:
+        assert args.conda_path
+
+    # remove any contents already present for a clean build
+    if os.path.exists(repo_dir):
+        shutil.rmtree(repo_dir)
+        os.makedirs(repo_dir)
+
+    # clone repo along with submodules
```
mikeroberts3000, line 36: Include ability to check out a specific commit

tools/package_executable.py
```diff
@@ -0,0 +1,129 @@
...
+    assert cmd_result.returncode == 0
+
+    # change working directory
+    cwd = os.getcwd()
+    print(f"[SPEAR | build.py] Changing working directory to {os.path.join(repo_dir, 'tools')}")
+    os.chdir(os.path.join(repo_dir, "tools"))
+
+    # build thirdparty libs
+    if sys.platform == "win32":
+        cmd = ["python", "build_third_party_libs.py", "--num_parallel_jobs", f"{args.num_parallel_jobs}"]
```
mikeroberts3000, line 49: Don't do this in a platform-specific way.

tools/package_executable.py
```diff
@@ -0,0 +1,129 @@
...
+    elif sys.platform in ["darwin", "linux"]:
+        cmd = [f"python build_third_party_libs.py --num_parallel_jobs {args.num_parallel_jobs}"]
+    else:
+        assert False
+    print(f"[SPEAR | build.py] Executing: {' '.join(cmd)}")
+    cmd_result = subprocess.run(cmd, shell=True)
+    assert cmd_result.returncode == 0
+
+    # create symbolic links
+    if sys.platform == "win32":
```
mikeroberts3000, line 59: No need for all of this platform checking.

tools/package_executable.py
```diff
@@ -0,0 +1,129 @@
...
+        cmd = [f". {args.conda_path}; conda activate {args.conda_env}; python generate_config.py"]
+    else:
+        assert False
+    print(f"[SPEAR | build.py] Executing: {' '.join(cmd)}")
+    cmd_result = subprocess.run(cmd, shell=True)
+    assert cmd_result.returncode == 0
+
+    # build the executable
+    if sys.platform == "win32":
+        build_exe = os.path.join(args.unreal_engine_dir, "Engine", "Build", "BatchFiles", "RunUAT.bat")
```
mikeroberts3000, line 82: run_uat_script

tools/package_executable.py
```diff
@@ -0,0 +1,129 @@
...
+
+    # clone repo along with submodules  
+    cmd = ["git", "clone", "--recurse-submodules", "https://github.com/isl-org/spear", repo_dir]
+    print(f"[SPEAR | build.py] Executing: {' '.join(cmd)}")
+    cmd_result = subprocess.run(cmd)
+    assert cmd_result.returncode == 0
+
+    # change working directory
+    cwd = os.getcwd()
+    print(f"[SPEAR | build.py] Changing working directory to {os.path.join(repo_dir, 'tools')}")
```
mikeroberts3000, line 44: Don't change the working directory here. Instead, just use the --unreal_project_dir command-line argument when calling create_sym_links.py and generate_config.py

tools/package_executable.py
```diff
@@ -0,0 +1,129 @@
...
+        platform_dir_name = "MacNoEditor"
+    elif sys.platform == "linux":
+        platform_dir_name = "LinuxNoEditor"
+    else:
+        assert False
+    
+    os.makedirs(args.output_dir, exist_ok=True)
+
+    # TODO: need to codesign on macOS
+    file = shutil.make_archive(base_name=os.path.join(args.output_dir, f"SpearSim-{args.output_tag}-{args.target_platform}-{args.config_mode}"), format='zip', root_dir=os.path.join(args.temp_dir, f"SpearSim-{args.config_mode}", platform_dir_name))
```
mikeroberts3000, line 126: - prefer double-quotes, e.g., format="zip"
- form the whole path in the sys.platform if statement, like you do for run_uat_script so that you don't need to call os.path.join inline when calling make_archive

tools/package_executable.py
```diff
@@ -0,0 +1,129 @@
...
+    shutil.rmtree(os.path.join(args.temp_dir, f"SpearSim-{args.config_mode}", platform_dir_name))
```
mikeroberts3000, line 127: - why are you removing this?
- when exactly would the "temp/SpearSim-Shipping/WindowsNoEditor" folder get created?

RachithP: Since we are removing packaging in this script, we no longer need this step

tools/package_executable.py
RachithP: Since we are removing packaging in this script, we no longer need this step

tools/package_executable.py
RachithP: It might be better to execute the scripts from the newly cloned dir because if there are any changes to these python scripts, the cloned repo will have the changes. I think it's cleaner to chdir() rather than calling commands like
```
cmd = ["python", os.path.join(repo_dir, "tools", "build_third_party_libs.py"), "--num_parallel_jobs", f"{args.num_parallel_jobs}"]
cmd = ["conda", "activate", args.conda_env + cmd_separator, "python", os.path.join(repo_dir, "tools", "create_symbolic_links.py"), "--unreal_project_dir", unreal_project_dir]
```

tools/package_executable.py
RachithP: updated

tools/package_executable.py
RachithP: updated

tools/package_executable.py
RachithP: updated

tools/package_executable.py
RachithP: updated

tools/package_executable.py
RachithP: Updated

tools/build.py
RachithP: Hmm, okay.

I'm handling the target_platform based on sys.platform, so we don't need `--target_platform` anymore

tools/build_executable.py
```diff
@@ -0,0 +1,105 @@
...
+    print(f"[SPEAR | build.py] Changing working directory to {os.path.join(repo_dir, 'tools')}")
+    os.chdir(os.path.join(repo_dir, "tools"))
+
+    if args.commit_id:
+        cmd = ["git", "reset", "--hard", args.commit_id]
+        print(f"[SPEAR | build.py] Executing: {' '.join(cmd)}")
+        cmd_result = subprocess.run(cmd)
+        assert cmd_result.returncode == 0
+
+    # build thirdparty libs
```
mikeroberts3000, line 63: third-party

tools/build_executable.py
```diff
@@ -0,0 +1,105 @@
...
+import argparse
+import os
+import shutil
+import subprocess
+import sys
+
+if __name__ == "__main__":
+
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--temp_dir", required=True)
```
mikeroberts3000, line 14: set this to be "tmp" as a default, or to os.path.join(os.path.dirname(os.path.realpath(__file__), "tmp") to match the existing convention for temp_dir in our code

tools/build_executable.py
```diff
@@ -0,0 +1,105 @@
...
+    cmd_result = subprocess.run(cmd, shell=True, executable=args.shell_bin)
+    assert cmd_result.returncode == 0
+
+    # generate config file
+    cmd = ["conda", "activate", args.conda_env + cmd_separator, "python", "generate_config.py"]
+    print(f"[SPEAR | build.py] Executing: {' '.join(cmd)}")
+    cmd_result = subprocess.run(cmd, shell=True, executable=args.shell_bin)
+    assert cmd_result.returncode == 0
+    
+    archive_dir = os.path.join(args.output_dir, f"SpearSim-{target_platform}-{args.config_mode}")
```
mikeroberts3000, line 81: place short descriptive comment here for consistency with the rest of the file

mikeroberts3000 (approved):
Looks great. Very clean and simple.

mikeroberts3000:
I just realized that the arguments are ordered as follows throughout our code: required args, args where we provide a default, optional args. We should order things in the same way in this latest batch of PRs.

tools/build_executable.py
```diff
@@ -0,0 +1,105 @@
...
+    ]
+    print(f"[SPEAR | build.py] Executing: {' '.join(cmd)}")
+    cmd_result = subprocess.run(cmd)
+    assert cmd_result.returncode == 0
+
+    # change working directory
+    print(f"[SPEAR | build.py] Changing working directory to {cwd}")
+    os.chdir(cwd)
+
+    print(f"[SPEAR | build.py] Successfully built SpearSim at {archive_dir}.")
```
mikeroberts3000, line 105: - do one last print statement after this to match the rest of our code: "[SPEAR | build_executable.py] Done."
- do the same thing in your other PRs

mikeroberts3000:
I think it would be less confusing if this script only executed Python code from the "outer" repo, i.e., not the repo that gets cloned at the beginning of `build_executable.py`.

Yes, the outer repo could potentially get out-of-sync with respect to the C++ code being built from the inner repo. But that is already an issue, regardless of which copy of `build_third_party_libs.py` we decide to execute. No matter what, when you `import spear`, even if you do so from `path/to/inner_repo/build_third_party_libs.py`, you'll be executing the code from the outer repo.

I think a simpler mental model to keep track of in our minds would be: "all python code comes from the outer repo; the only thing that comes from the inner repo is the C++ code being built"

In contrast, the current mental model implied by this PR is more complicated: "`build_executable.py` is from the outer repo; the python scripts it invokes like `build_third_party_libs.py` are from the inner repo; `import spear` and `spear.get_config()` are from the outer repo; the C++ code is from the inner repo"

RachithP:
Hmmm, I think it's a bit more complicated/tricky. If we execute all python scripts from "outer" repo, our `create_symbolic_links.py` script links `unreal_plugins` from the "outer" repo to the "inner" repo's `unreal_projects`. This would be mixing `C++` code from "outer" and "inner" repo. This shouldn't happen.

One way I can think of is to modify `create_symbolic_links.py` to take `unreal_plugins_dir`, `third_party_dir` as inputs in addition to `unreal_project_dir`, and set defaults as in the current script. This way, we avoid this mixing of `C++` code.

tools/build_executable.py
RachithP: Done

tools/build_executable.py
RachithP: Done

tools/build_executable.py
RachithP: Done

tools/build_executable.py
RachithP: Done

mikeroberts3000:
I like that solution.

tools/build_executable.py
```diff
@@ -0,0 +1,108 @@
...
+        target_platform = "Win64"
+        archive_dir = os.path.join(args.output_dir, f"SpearSim-{target_platform}-{args.config_mode}")
+        cmd_prefix = f"conda activate {args.conda_env}&"
+    elif sys.platform == "darwin":
+        run_uat_script = os.path.join(args.unreal_engine_dir, "Engine", "Build", "BatchFiles", "RunUAT.sh")
+        target_platform = "Mac"
+        archive_dir = os.path.join(args.output_dir, f"SpearSim-{target_platform}-{args.config_mode}-Unsigned")
+        conda_script = args.conda_script
+        if not conda_script:
+            conda_script = "~/anaconda3/etc/profile.d/conda.sh"
```
mikeroberts3000, line 44: `os.path.join( "~", "anaconda3", "etc", "profile.d", "conda.sh")`

Strictly speaking, you don't need to do this, but it's just for consistency with `run_uat_script`.

tools/build_executable.py
```diff
@@ -0,0 +1,108 @@
...
+        cmd_prefix = f". {conda_script}; conda activate {args.conda_env};"
```
mikeroberts3000, line 45: f". {conda_script}; conda activate {args.conda_env};SPACE"

tools/build_executable.py
```diff
@@ -0,0 +1,108 @@
...
+    assert cmd_result.returncode == 0
+
+    if args.commit_id:
+        cmd = ["git", "reset", "--hard", args.commit_id]
+        print(f"[SPEAR | build_executable.py] Executing: {' '.join(cmd)}")
+        cmd_result = subprocess.run(cmd, check=True)
+        assert cmd_result.returncode == 0
+
+    # build third-party libs
+    cmd = ["python", "build_third_party_libs.py", "--num_parallel_jobs", f"{args.num_parallel_jobs}"]
```
mikeroberts3000, line 70: Why are we storing cmd as a list sometimes and as a string other times? Whatever we do in this file, we should be consistent.

tools/build_executable.py
```diff
@@ -0,0 +1,108 @@
...
+    print(f"[SPEAR | build_executable.py] Executing: {' '.join(cmd)}")
+    cmd_result = subprocess.run(cmd, check=True)
+    assert cmd_result.returncode == 0
+
+    # create symbolic links
+    cmd = cmd_prefix + f"python create_symbolic_links.py --unreal_project_dir {unreal_project_dir} --unreal_plugins_dir {unreal_plugins_dir}"
+    print(f"[SPEAR | build_executable.py] Executing: {cmd}")
+    cmd_result = subprocess.run(cmd, shell=True, check=True)
+    assert cmd_result.returncode == 0
```
mikeroberts3000, line 79: if check=True, do we need assert cmd_result.returncode == 0?

tools/create_symbolic_links.py
```diff
@@ -24,6 +26,12 @@
     else:
         unreal_project_dirs = [ os.path.join(unreal_projects_dir, project) for project in os.listdir(unreal_projects_dir) ]
 
+    if args.unreal_plugins_dir:
```
mikeroberts3000, line 29: Don't you need else statements here? What happens if I don't specify args.unreal_plugins_dir?

mikeroberts3000:
I made a few notes but otherwise this looks great.

tools/create_symbolic_links.py
RachithP: We don't because we are writing default values to this variable few lines above. If args.unreal_plugins_dir is input, we overwrite those default values.

tools/build_executable.py
RachithP: We don't need the assert statement. I removed it.

tools/build_executable.py
RachithP: Done

tools/build_executable.py
RachithP: Done

tools/build_executable.py
RachithP: When we execute with `shell=True`, it needs to be a string, otherwise it should be a list.

mikeroberts3000: Ah ok got it.

tools/create_symbolic_links.py
mikeroberts3000: Ah ok. I prefer if-else for this kind of thing.

```
if args.unreal_plugins_dir:
    unreal_plugins_dir = args.unreal_plugins_dir
else:
    unreal_plugins_dir = whatever
```

Because it localizes the conditional logic to a very small part of the code. Otherwise, I need to scan through the code for a potentially unlimited number of lines to figure out the conditional logic.

mikeroberts3000:
Note: since we removed the creation of the final zip file out of `build_executable.py`, we need to create a simple `archive_executable.py` script that will create the final zip file for each platform. We might need to call `ditto` on macOS in case that is needed for some reason, but I'm not sure.

tools/create_symbolic_links.py
RachithP: I agree, it's more clear this way. I see that you have made these changes. Thanks!

RachithP:
> Note: since we removed the creation of the final zip file out of `build_executable.py`, we need to create a simple `archive_executable.py` script that will create the final zip file for each platform. We might need to call `ditto` on macOS in case that is needed for some reason, but I'm not sure.

Yes, I'll create this script.

PR #157, opened 2023-03-01
UrdfBot Joint Torque based  Control

rwang15:
Stability and correctness checked in branch [rwang15/task-space-control](https://github.com/isl-org/spear/tree/rwang15/task-space-control).


rwang15:
Staying with previous method for parsing urdf, i.e. mirrored appearance compared to Mujoco, while rotation might not be parsed correctly.

PR #158, opened 2023-03-03
cleaned up tools folder

mikeroberts3000:
- use the inner `third_party` folder when building an executable via `build_executable.py` (the current code uses the outer `third_party` folder, but it is more consistent and simpler approach to use the inner `third_party` folder, because then all C++ code comes from the inner repo and all Python code comes from the outer repo)
- fixed a bug where cooking warnings occur during normal builds after executing `build_executable.py` on macOS
- hard-code all build tools to assume a single `SpearSim` project in `unreal_projects` for consistency, i.e., don't attempt to iterate over all projects in `unreal_projects`
- `os.path.realpath` is always the left-most operator for readability and consistency


tools/build_executable.py
```diff
@@ -8,85 +8,107 @@
...
+        shutil.rmtree(repo_dir)
+        os.makedirs(repo_dir)
+
+    # We need to remove this temp dir (created by the Unreal build process) because it contains paths from previous builds.
+    # If we don't do this step, we will get many warnings during this build:
+    #     Warning: Unable to generate long package name for path/to/previous/build/Some.uasset because FilenameToLongPackageName failed to convert
+    #     'path/to/previous/build/Some.uasset'. Attempt result was '../../../../../../path/to/previous/build/path/to/previous/build/Some', but the
+    #     path contains illegal characters '.'
+    print(unreal_tmp_dir)
+    print()
```
RachithP, line 81: Remove these print statements.

tools/build_executable.py
```diff
@@ -8,85 +8,107 @@
...
+    if os.path.exists(unreal_tmp_dir):
+        print(f"[SPEAR | build_executable.py] Global cache directory exists, removing: {unreal_tmp_dir}")
```
RachithP, line 83: Should we add `unreal engine` to the print statement? This way it makes it more clear that the global cache directory we reference here is managed by Unreal. Something like
```python
print(f"[SPEAR | build_executable.py] Unreal Engine's global cache directory exists, removing: {unreal_tmp_dir}")
```

tools/create_symbolic_links.py
```diff
@@ -11,28 +11,14 @@
...
-    else:
-        unreal_plugins_dir = os.path.realpath(os.path.join(tools_dir, "..", "cpp", "unreal_plugins"))
-
-    if args.third_party_dir:
-        third_party_dir = os.path.realpath(args.third_party_dir)
-    else:
-        third_party_dir = os.path.realpath(os.path.join(tools_dir, "..", "third_party"))
+    unreal_project_dir = os.path.realpath(args.unreal_project_dir)
+    unreal_plugins_dir = os.path.realpath(args.unreal_plugins_dir)
+    third_party_dir = os.path.realpath(args.third_party_dir)
```
RachithP, line 21: Let's vertically align `=`

tools/sign_executable/sign_executable.py
```diff
@@ -19,13 +19,13 @@
...
 
     parser = argparse.ArgumentParser()
-    parser.add_argument("--input_dir", required=True)
-    parser.add_argument("--output_dir", required=True)
     parser.add_argument("--developer_id", required=True)
     parser.add_argument("--apple_username", required=True)
     parser.add_argument("--apple_password", required=True)
-    parser.add_argument("--temp_dir", default=os.path.join(os.path.dirname(os.path.realpath(__file__)), "tmp"))
-    parser.add_argument("--entitlements_file", default=os.path.join(os.path.dirname(os.path.realpath(__file__)), "entitlements.plist"))
+    parser.add_argument("--input_dir", default=os.path.realpath(os.path.join(os.path.dirname(__file__), "tmp", "SpearSim-Mac-Shipping-Unsigned")))
```
RachithP, line 25: The default `input_dir` in `sign_executable.py` and the default `output_dir` in `build_executable.py` do not match. They need to be matched for this script to find `SpearSim-Mac-Shipping-Unsigned` in default settings.

RachithP (approved):
Some minor comments. Rest looks good.

.gitignore
```diff
@@ -54,3 +54,4 @@ examples/imitation_learning_openbot/videos
 python/spear_sim.egg-info
 
 tools/scene_downloader/tmp
+tools/tmp
```
RachithP, line 57: I think we should also add `tools/sign_executable/tmp`, since we are creating this temporary folder in our `sign_executable.py` script.

tools/build_executable.py
mikeroberts3000: Good catch, removed.

tools/build_executable.py
mikeroberts3000: Agreed.

tools/create_symbolic_links.py
mikeroberts3000: Sure.

tools/sign_executable/sign_executable.py
mikeroberts3000: I agree that the default values in `build_executable` and `sign_executable` should match. I changed the logic slightly in `sign_executable` in a way that I think is better, such that: (1) the default values match `build_executable`; and (2) the output folder matches the structure of the `-archive` step in the Unreal Build Tool, i.e., `sign_executable` doesn't modify the existing folder structure at all.

.gitignore
mikeroberts3000: Agreed, although `sign_executable` now defaults to using `tools/tmp` as its `temp_dir`, because I wanted it to match the default `temp_dir` used by `build_executable`. But there is no harm in adding `tools/sign_executable/tmp` to the ignore list, and doing so would be consistent with `scene_downloader`.

PR #161, opened 2023-03-09
add config flag to support non-config mode

rwang15:
add config flag to allow fly around the scene when execute standalone-executable without `config.yaml`.
* add Config flag checking if `config.yaml` is provided;
* skip UObject constructor if it depends on Config;
* custom default GameMode with default Pawn of `ASpectatorPawn`.

### DIFF
--- a/cpp/unreal_plugins/CoreUtils/Source/CoreUtils/Config.cpp
+++ b/cpp/unreal_plugins/CoreUtils/Source/CoreUtils/Config.cpp
@@ -12,6 +12,7 @@
 #include "CoreUtils/Unreal.h"
 #include "CoreUtils/YamlCpp.h"
 
+bool Config::is_enabled_;
 YAML::Node Config::s_config_;
 
 void Config::initialize()
@@ -21,19 +22,26 @@ void Config::initialize()
     // If a config file is provided via the command-line, then load it
     if (FParse::Value(FCommandLine::Get(), TEXT("config_file="), config_file)) {
         s_config_ = YAML::LoadFile(Unreal::toStdString(config_file));
+        is_enabled_=true;
 
     // Otherwise, if MyProject/Temp/config.yaml exists, then load it
     } else if (FPaths::FileExists(FPaths::ConvertRelativePathToFull(FPaths::ProjectDir().Append("Temp/config.yaml")))) {
         config_file = FPaths::ConvertRelativePathToFull(FPaths::ProjectDir().Append("Temp/config.yaml"));
         s_config_ = YAML::LoadFile(Unreal::toStdString(config_file));
+        is_enabled_=true;
 
     // Otherwise assert
     } else {
-        ASSERT(false);
+        is_enabled_ = false;
     }
 }
 
 void Config::terminate()
 {
     s_config_.reset();
+}
+
+bool Config::isEnabled()
+{
+    return is_enabled_;
 }
--- a/cpp/unreal_plugins/CoreUtils/Source/CoreUtils/Config.h
+++ b/cpp/unreal_plugins/CoreUtils/Source/CoreUtils/Config.h
@@ -17,6 +17,7 @@ class COREUTILS_API Config
 public:
     static void initialize();
     static void terminate();
+    static bool isEnabled();
 
     //
     // This function is used to extract a value from the Config system. This function takes as input
@@ -73,5 +74,6 @@ private:
     Config() = default;
     ~Config() = default;
 
+    static bool is_enabled_;
     static YAML::Node s_config_;
 };
--- a/cpp/unreal_plugins/CoreUtils/Source/CoreUtils/CoreUtils.cpp
+++ b/cpp/unreal_plugins/CoreUtils/Source/CoreUtils/CoreUtils.cpp
@@ -15,7 +15,7 @@ void CoreUtils::StartupModule()
     Config::initialize();
 
     // Wait for keyboard input, which is useful when attempting to attach a debugger to the running executable.
-    if (Config::get<bool>("CORE_UTILS.WAIT_FOR_KEYBOARD_INPUT_DURING_INITIALIZATION")) {
+    if (Config::isEnabled() && Config::get<bool>("CORE_UTILS.WAIT_FOR_KEYBOARD_INPUT_DURING_INITIALIZATION")) {
         std::cout << "[SPEAR | CoreUtils.cpp] Press ENTER to continue..." << std::endl;
         std::cin.get();
     }
--- a/cpp/unreal_plugins/SimulationController/Source/SimulationController/SimulationController.cpp
+++ b/cpp/unreal_plugins/SimulationController/Source/SimulationController/SimulationController.cpp
@@ -90,7 +90,7 @@ void SimulationController::postWorldInitializationEventHandler(UWorld* world, co
 
     ASSERT(world);
 
-    if (world->IsGameWorld() && GEngine->GetWorldContextFromWorld(world)) {
+    if (Config::isEnabled() && world->IsGameWorld() && GEngine->GetWorldContextFromWorld(world)) {
         auto world_path_name = Config::get<std::string>("SIMULATION_CONTROLLER.WORLD_PATH_NAME");
         auto level_name = Config::get<std::string>("SIMULATION_CONTROLLER.LEVEL_NAME");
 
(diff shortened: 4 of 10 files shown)


cpp/unreal_plugins/CoreUtils/Source/CoreUtils/Config.cpp
```diff
@@ -12,6 +12,7 @@
 #include "CoreUtils/Unreal.h"
 #include "CoreUtils/YamlCpp.h"
 
+bool Config::is_enabled_;
```
mikeroberts3000, line 15: I would rather keep this logic out of `Config`. Instead, I think it would be better to define a public enum variable directly in the `CoreUtils` class.

```
enum class LaunchMode
{
    Default,
    Python,
}

class CoreUtils : public IModuleInterface
{
public:
    void StartupModule() override;
    void ShutdownModule() override;

    static LaunchMode s_launch_mode_;
};
```

You could set this variable as follows.

```
CoreUtils::s_launch_mode_ = LaunchMode::Default;

void CoreUtils::StartupModule()
{
    std::cout << "[SPEAR | CoreUtils.cpp] CoreUtils::StartupModule" << std::endl;

    // set launch mode
    FString launch_mode;
    if (FParse::Value(FCommandLine::Get(), TEXT("launch_mode="), launch_mode)) {
        if (Unreal::toStdString(launch_mode) == "python") {
            s_launch_mode_ = LaunchMode::Python;
        } else {
            ASSERT(false);
        }
    }

    // if we weren't launched via Python, then return
    if (s_launch_mode_ == LaunchMode::Default) {
        return;
    }

    Config::initialize();

    // ...
}
```

I like this design better because it expresses our intent more clearly throughout our code. For example, I could imagine wanting to disable a `FastCppPythonSharedMemory` system, or something else that doesn't depend on our config system at all, when I'm in `LaunchMode::Default`. If `FastCppPythonSharedMemory` never uses our config system, it would be confusing for it to check `if (Config::isEnabled()) return`. In this case, it would be more clear if `FastCppPythonSharedMemory` checked `if (CoreUtils::s_launch_mode_ == LaunchMode::Default) return` directly.

cpp/unreal_plugins/CoreUtils/Source/CoreUtils/Config.cpp
```diff
@@ -21,19 +22,26 @@ void Config::initialize()
     // If a config file is provided via the command-line, then load it
     if (FParse::Value(FCommandLine::Get(), TEXT("config_file="), config_file)) {
         s_config_ = YAML::LoadFile(Unreal::toStdString(config_file));
+        is_enabled_=true;
 
     // Otherwise, if MyProject/Temp/config.yaml exists, then load it
```
mikeroberts3000, line 27: I think we can get rid of this `else if` statement, and we can also remove `generate_config.py` from our build steps. Nice!

cpp/unreal_plugins/SimulationController/Source/SimulationController/SimulationController.cpp
```diff
@@ -90,7 +90,7 @@ void SimulationController::postWorldInitializationEventHandler(UWorld* world, co
 
     ASSERT(world);
 
-    if (world->IsGameWorld() && GEngine->GetWorldContextFromWorld(world)) {
+    if (Config::isEnabled() && world->IsGameWorld() && GEngine->GetWorldContextFromWorld(world)) {
```
mikeroberts3000, line 93: Better to check in `SimulationController::StartupModule()`. From there, you can prevent everything in `SimulationController` from being executed in a very clean way.

```
void SimulationController::StartupModule()
{
    std::cout << "[SPEAR | SimulationController.cpp] SimulationController::StartupModule" << std::endl;

    ASSERT(FModuleManager::Get().IsModuleLoaded(TEXT("CoreUtils")));
    ASSERT(FModuleManager::Get().IsModuleLoaded(TEXT("OpenBot")));
    ASSERT(FModuleManager::Get().IsModuleLoaded(TEXT("UrdfBot")));

    if (CoreUtils::s_launch_mode_ == LaunchMode::Default) {
        return;
    }
    
    // ...
}

void SimulationController::ShutdownModule()
{
    std::cout << "[SPEAR | SimulationController.cpp] SimulationController::ShutdownModule" << std::endl;

    if (CoreUtils::s_launch_mode_ == LaunchMode::Default) {
        return;
    }
    
    // ...
}
```

cpp/unreal_projects/SpearSim/Config/DefaultEngine.ini
```diff
@@ -29,6 +29,7 @@ r.Streaming.FullyLoadUsedTextures=1
 [/Script/EngineSettings.GameMapsSettings]
 GameDefaultMap=/Game/Scenes/default/Maps/DefaultMap.DefaultMap
 EditorStartupMap=/Game/Scenes/default/Maps/DefaultMap.DefaultMap
+GlobalDefaultGameMode=/Script/SpearSim.SpearSimGameMode
```
mikeroberts3000, line 32: Nice. I wonder how difficult it is to make `ASpectatorPawn` receive input and update the viewport while the game is paused. If this is easy to do, then we would effectively a nice debug camera that could move freely even when our simulation is paused and we're waiting for Python input.

This functionality already works in Debug builds by activating the debug camera through the debug console, but the default debug camera is too sensitive to keyboard input so it is hard to navigate, and it would be nice to enable this functionality on Shipping builds.

cpp/unreal_projects/SpearSim/Source/SpearSim/SpearSimGameMode.h
```diff
@@ -0,0 +1,22 @@
...
+//
+
+#pragma once
+
+#include "CoreMinimal.h"
+#include "GameFramework/GameModeBase.h"
+
+#include "SpearSimGameMode.generated.h"
+
+/**
```
mikeroberts3000, line 13: Remove this style of comment.

cpp/unreal_plugins/CoreUtils/Source/CoreUtils/Config.cpp
mikeroberts3000: Note that you'll need to add `-launch_mode=python` to the command-line arguments in `env.py`.

mikeroberts3000: I'm not sure if you're allowed to use the `COREUTILS_API` macro in this way. And maybe you'll need to apply it to the enum class also? If doing all of this directly in the `CoreUtils` class doesn't work, you could make a new class called `LaunchMode` that handles all of the launch mode functionality.

mikeroberts3000, line 15: Update: based on an offline design discussion, we agreed that it is better to go with the sketch as it is proposed in this PR. But let's use `bool Config::s_is_enabled_` to match our existing convention for static variables.

cpp/unreal_plugins/CoreUtils/Source/CoreUtils/Config.cpp
mikeroberts3000: Update: we can get rid of this `else if` statement but not `generate_config.py`.

cpp/unreal_plugins/CoreUtils/Source/CoreUtils/Config.h
```diff
@@ -17,6 +17,7 @@ class COREUTILS_API Config
 public:
     static void initialize();
     static void terminate();
+    static bool isEnabled();
```
mikeroberts3000, line 20: Update: no need for this trivial get function, we can just expose the bool directly as a public variable.

cpp/unreal_plugins/CoreUtils/Source/CoreUtils/CoreUtils.cpp
```diff
@@ -15,7 +15,7 @@ void CoreUtils::StartupModule()
     Config::initialize();
 
     // Wait for keyboard input, which is useful when attempting to attach a debugger to the running executable.
-    if (Config::get<bool>("CORE_UTILS.WAIT_FOR_KEYBOARD_INPUT_DURING_INITIALIZATION")) {
+    if (Config::isEnabled() && Config::get<bool>("CORE_UTILS.WAIT_FOR_KEYBOARD_INPUT_DURING_INITIALIZATION")) {
```
mikeroberts3000, line 18: Better to structure this check as an early return statement, because then it will match the other code in our all of the constructors etc.

mikeroberts3000 (approved):

PR #164, opened 2023-03-15
Updated IL navigation model

quentin-leboutet:
Updated the navigation model in the imitation learning example.


mikeroberts3000 (approved):
Approved. Yay!

PR #168, opened 2023-03-24
improved behavior when double-clicking on the SpearSim executable

mikeroberts3000:
improved behavior when double-clicking on the SpearSim executable: runs in windowed mode, and the user can press escape to exit out of the application


RachithP (approved):
Looks good!

PR #170, opened 2023-03-24
 Update build script to use new pak file naming convention

RachithP:


mikeroberts3000:
The arguments here are called "tag" in archive_executable.py, "paks_version" in build_executable.py, and "version_tag" in build_pak_files.py. This is too inconsistent. It should be called the same thing everywhere. I like "version_tag".

RachithP:
> The relevant arguments here are called `tag` in `archive_executable.py`, `paks_version` in `build_executable.py`, and `version_tag` in `build_pak_files.py`. This is too inconsistent. It should be called the same thing everywhere. I like `version_tag`.

I agree. Updated the argument names.

mikeroberts3000 (approved):

PR #171, opened 2023-03-25
fix build on Windows and update `sign_executable.py` 

RachithP:


RachithP:
Included the changes here in https://github.com/isl-org/spear/pull/172.

PR #172, opened 2023-03-25
Windows build fix + update build scripts to copy paks files during archive process

RachithP:
- Fix build issue on Windows platform. Add `InputCore` module dependency to `SpearSim.Build.cs`.
- Copying our default pak file (of map `kujiale_0000`) into the executable as part of `archive_executable.py` instead of `build_executable.py`.
  - The reason I'm suggesting this change is that signing mac executable takes a very long time due to the large executable size as it needs to upload the executable for notarization.
  - Since the pak file copying step is needed for public release builds, it should be fine if it's part of `archive_executable.py` instead of `build_executable.py` as `archive_executable.py` is required to be used for the public release as well.

### DIFF
--- a/tools/build_executable/build_executable.py
+++ b/tools/build_executable/build_executable.py
@@ -14,6 +14,7 @@ if __name__ == "__main__":
     parser = argparse.ArgumentParser()
     parser.add_argument("--unreal_engine_dir", required=True)
     parser.add_argument("--paks_dir", required=True)
+    parser.add_argument("--version_tag", required=True)
     parser.add_argument("--conda_env", default="spear-env")
     parser.add_argument("--num_parallel_jobs", type=int, default=1)
     parser.add_argument("--output_dir", default=os.path.realpath(os.path.join(os.path.dirname(__file__), "..", "tmp")))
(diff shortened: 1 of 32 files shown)


tools/build_executable/build_executable.py
```diff
@@ -13,8 +13,6 @@
 
     parser = argparse.ArgumentParser()
     parser.add_argument("--unreal_engine_dir", required=True)
-    parser.add_argument("--paks_dir", required=True)
```
mikeroberts3000, line 16: I realize I've gone back and forth on this issue, but after reviewing this code more closely, I think the approach in this PR is flawed.

In this PR, I don't like how `build_executable` will produce an executable that is in a broken non-working state. And I don't like how `archive_executable` does more than generate an archive, i.e., it has some hidden responsibilities that aren't reflected in its name.

In our existing approach, I like the fact that I can run `build_executable` and then get an executable that is usable. I shouldn't have to run `archive_executable` to get a zip file, and then unzip it, just to get back to a usable executable. And, conceptually, I like the fact that `build_executable` handles the all the responsibilities of setting the default map (changing GameDefaultMap and copying the pak file), rather than having the responsibilities scattered across two different files.

I'd like to propose an alternative approach. We should keep the full functionality in `build_executable`. In `sign_executable`, we should delete the pak file from the working copy that we're modifying anyway, then upload for signing, then re-copy the pak file into the working copy. At first glance, this approach seems silly because it involves one file's worth of redundant copying, but it has the following advantages: (1) the output of every script is fully functional instead of broken; (2) the responsibilities of each script are more logically delineated; (3) we still keep the advantages of this PR, namely we don't need to upload huge executables for signing.

RachithP: Your suggested idea is definitely better than the current version. I'll update the scripts accordingly!

mikeroberts3000 (approved):
Great.

PR #173, opened 2023-03-26
implemented run_executable.py and other improvements 

mikeroberts3000:
- implemented `tools/run_executable.py`, which can be used to launch a shipping executable, load a scene specified on the command-line, and let the user navigate around the scene interactively using the keyboard
- this functionality needed to be implemented in a standalone Python tool because it must create a symlink from the `Contents` directory of the executable to a user's local pak directory, and this must be done prior to launching the executable
- it is also not possible to specify a map on the command-line for shipping builds (without rebuilding the engine with a specific preprocessor macro defined), so we need some kind of wrapper around a shipping executable that enables loading a specific map via the command-line
- improved configuration options for loading a specific scene; now the user specifies `SIMULATION_CONTROLLER.SCENE_ID` and optionally `SIMULATION_CONTROLLER.MAP_ID`, which will load `/Game/Scenes/SCENE_ID/Maps/MAP_ID.MAP_ID`; if `MAP_ID` is blank, it defaults to the same value as `SCENE_ID`
- made the use of `os.path.realpath` more consistent and more readable throughout the codebase
- removed the `SPEAR.CONTENT_DIR` configuration option because it can be inferred automatically from the executable path
- added a build step for boost that simplifies how we include boost packages in our `Build.cs` files, and will be required if we ever want to use a boost package that is not header-only (e.g., `boost::filesystem`)

### DIFF
--- a/python/spear/env.py
+++ b/python/spear/env.py
@@ -100,20 +100,33 @@ class Env(gym.Env):
             return
 
         # write temp file
-        temp_config_file = os.path.realpath(os.path.join(self._config.SPEAR.TEMP_DIR, "config.yaml"))
+        temp_dir = os.path.realpath(os.path.join(self._config.SPEAR.TEMP_DIR))
+        temp_config_file = os.path.realpath(os.path.join(temp_dir, "config.yaml"))
 
         print("[SPEAR | env.py] Writing temp config file: " + temp_config_file)
 
-        if not os.path.exists(os.path.realpath(self._config.SPEAR.TEMP_DIR)):
-            os.makedirs(os.path.realpath(self._config.SPEAR.TEMP_DIR))
+        if not os.path.exists(temp_dir):
+            os.makedirs(temp_dir)
         with open(temp_config_file, "w") as output:
             self._config.dump(stream=output, default_flow_style=False)
 
         # create a symlink to SPEAR.DATA_DIR
         if self._config.SPEAR.DATA_DIR != "":
 
-            paks_dir = os.path.join(self._config.SPEAR.CONTENT_DIR, "Paks")
+            assert os.path.exists(self._config.SPEAR.DATA_DIR)
+
+            if sys.platform == "win32":
+                paks_dir = os.path.realpath(os.path.join(os.path.dirname(os.path.realpath(args.executable)), "..", "..", "Content", "Paks"))
+            elif sys.platform == "darwin":
+                paks_dir = os.path.realpath(os.path.join(args.executable, "Contents", "UE4", "SpearSim", "Content", "Paks"))
+            elif sys.platform == "linux":
+                paks_dir = os.path.realpath(os.path.join(os.path.dirname(os.path.realpath(args.executable)), "SpearSim", "Content", "Paks"))
+            else:
+                assert False
+
             assert os.path.exists(paks_dir)
+
+            # we don't use os.path.realpath here because we don't want to resolve the symlink
             spear_data_dir = os.path.join(paks_dir, "SpearData")
 
             if spear.path_exists(spear_data_dir):
@@ -121,7 +134,7 @@ class Env(gym.Env):
                 spear.remove_path(spear_data_dir)
 
             print(f"[SPEAR | env.py] Creating symlink: {spear_data_dir} -> {self._config.SPEAR.DATA_DIR}")
-            os.symlink(self._config.SPEAR.DATA_DIR, spear_data_dir)
+            os.symlink(args.data_dir, spear_data_dir)
 
         # provide additional control over which Vulkan devices are recognized by Unreal
         if len(self._config.SPEAR.VULKAN_DEVICE_FILES) > 0:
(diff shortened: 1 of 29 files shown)


tools/build_third_party_libs.py
```diff
@@ -51,6 +51,84 @@
...
 
+    #
+    # Boost
+    #
+
+    print("[SPEAR | build_third_party_libs.py] Building Boost...")
+
+    boost_dir = os.path.realpath(os.path.join(third_party_dir, "boost"))
+    include_dir = os.path.realpath(os.path.join(third_party_dir, "boost", "boost"))
+    build_dir = os.path.realpath(os.path.join(third_party_dir, "boost", "stage"))
```
RachithP, line 62: Let's vertically align these statements.

tools/build_third_party_libs.py
```diff
@@ -51,6 +51,84 @@
...
+
+    if os.path.isdir(include_dir):
+        print(f"[SPEAR | build_third_party_libs.py] Directory exists, removing: {include_dir}")
+        shutil.rmtree(include_dir, ignore_errors=True)
+
+    if os.path.isdir(build_dir):
+        print(f"[SPEAR | build_third_party_libs.py] Directory exists, removing: {include_dir}")
+        shutil.rmtree(include_dir, ignore_errors=True)
+
+    print(f"[SPEAR | build_third_party_libs.py] Creating directory and changing to working: {boost_dir}")
```
RachithP, line 81: We are only changing paths. Need to update the print statement.

python/spear/env.py
```diff
@@ -100,28 +100,41 @@ def _request_launch_unreal_instance(self):
...
             self._config.dump(stream=output, default_flow_style=False)
 
         # create a symlink to SPEAR.DATA_DIR
         if self._config.SPEAR.DATA_DIR != "":
 
-            paks_dir = os.path.join(self._config.SPEAR.CONTENT_DIR, "Paks")
+            assert os.path.exists(self._config.SPEAR.DATA_DIR)
+
+            if sys.platform == "win32":
+                paks_dir = os.path.realpath(os.path.join(os.path.dirname(os.path.realpath(args.executable)), "..", "..", "Content", "Paks"))
```
RachithP, line 119: `args.executable` is not valid here. Similarly, `args.data_dir`.

RachithP: I think we should change it to `self._config.SPEAR.STANDALONE_EXECUTABLE` and add another condition in the if statement of #114,
```python 
if self._config.SPEAR.STANDALONE_EXECUTABLE != "" and self._config.SPEAR.DATA_DIR != "":
```

RachithP: If we need to support symlink creation for both editor mode and standalone mode, then we will have to handle these paths differently.

python/spear/env.py
```diff
@@ -100,28 +100,41 @@ def _request_launch_unreal_instance(self):
...
+            # we don't use os.path.realpath here because we don't want to resolve the symlink
             spear_data_dir = os.path.join(paks_dir, "SpearData")
 
             if spear.path_exists(spear_data_dir):
                 print(f"[SPEAR | env.py] File or directory or symlink exists, removing: {spear_data_dir}")
                 spear.remove_path(spear_data_dir)
 
             print(f"[SPEAR | env.py] Creating symlink: {spear_data_dir} -> {self._config.SPEAR.DATA_DIR}")
-            os.symlink(self._config.SPEAR.DATA_DIR, spear_data_dir)
+            os.symlink(args.data_dir, spear_data_dir)
```
RachithP, line 137: This should be `self._config.SPEAR.DATA_DIR`, not `args.data_dir`.

tools/run_executable.py
```diff
@@ -0,0 +1,132 @@
...
+    
+    assert os.path.exists(args.executable)
+
+    # determine the internal executable we will actually launch
+    executable_name, executable_ext = os.path.splitext(args.executable)
+
+    if sys.platform == "win32":
+        assert executable_name[-4:] == "-Cmd"
+        assert executable_ext == ".exe"
+        executable_internal = executable
```
RachithP, line 30:
```python
executable_internal = args.executable
```

tools/run_executable.py
```diff
@@ -0,0 +1,132 @@
...
+        launch_args.append("{}".format(a))
+
+    # launch executable
+    cmd = [executable_internal] + launch_args
+
+    print("[SPEAR | run_executable.py] Launching executable with the following command-line arguments:")
+    print(" ".join(cmd))
+
+    print("[SPEAR | run_executable.py] Launching executable with the following config values:")
+    print(config)
```
RachithP, line 128: I think it will be better if we interchange these two sets of print statements. Printing config will be long and will mask the previous print statement, and I think having the launch command as the last print statement will be more helpful.

tools/build_third_party_libs.py
mikeroberts3000: I agree with you, although this is a matter of taste, and our current convention does not mandate that we align equal signs in all cases. But in this case, it clearly makes the code more beautiful.

tools/build_third_party_libs.py
mikeroberts3000: Good catch.

python/spear/env.py
mikeroberts3000: Good catch. I think if you're running in editor mode for some reason, then you don't need pak files. So we only need to do this step in standalone mode. I updated the logic accordingly.

mikeroberts3000: I changed it to:

```
        if self._config.SPEAR.LAUNCH_MODE == "standalone_executable" and self._config.SPEAR.DATA_DIR != "":

            assert os.path.exists(self._config.SPEAR.STANDALONE_EXECUTABLE)
            assert os.path.exists(self._config.SPEAR.DATA_DIR)
```

python/spear/env.py
mikeroberts3000: Oof. That's embarrassing. Good catch.

tools/run_executable.py
mikeroberts3000: Good catch.

tools/run_executable.py
mikeroberts3000: I somewhat disagree, and I'm maintaining consistency with what we already do in env.py

RachithP (approved):
Requires few changes, everything else looks good!

PR #175, opened 2023-03-27
tool for creating and removing symlinks to a specific scene

mikeroberts3000:
this tool enables editing scenes in the Unreal Editor, via SpearSim.uproject, without needing to copy folders back and forth from the user's local Perforce workspace

### DIFF
--- a/tools/prepare_scene_for_editing.py
+++ b/tools/prepare_scene_for_editing.py
@@ -40,7 +40,6 @@ if __name__ == '__main__':
         os.symlink(perforce_content_shared_dir, unreal_project_content_shared_dir)
 
     # Create and/or remove symlink for scene directory
-
     perforce_content_scene_dir = os.path.realpath(os.path.join(args.perforce_content_dir, "Scenes", args.scene_id))
 
     # We do not want to use os.path.realpath(...) here, because that will resolve to the directory inside the user's Perforce workspace.
--- a/tools/prepare_scene_for_editing.py
+++ b/tools/prepare_scene_for_editing.py
@@ -1,0 +1,60 @@
+#
+# Copyright(c) 2022 Intel. Licensed under the MIT License <http://opensource.org/licenses/MIT>.
+#
+
+import argparse
+import fnmatch
+import glob
+import os
+import posixpath
+import spear
+import subprocess
+import sys
+
+
+if __name__ == '__main__':
+
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--perforce_content_dir", required=True)
+    parser.add_argument("--scene_id", required=True)
+    parser.add_argument("--remove_symlink", action="store_true")
+    args = parser.parse_args()
+    
+    assert os.path.exists(args.perforce_content_dir)
+
+    unreal_project_content_dir = os.path.realpath(os.path.join(os.path.dirname(__file__), "..", "cpp", "unreal_projects", "SpearSim", "Content"))
+
+    # Create and/or remove symlink for Shared directory
+    perforce_content_shared_dir = os.path.realpath(os.path.join(args.perforce_content_dir, "Shared"))
+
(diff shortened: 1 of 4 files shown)


tools/prepare_scene_for_editing.py
```diff
@@ -0,0 +1,54 @@
...
+    parser.add_argument("--remove_symlink", action="store_true")
+    args = parser.parse_args()
+    
+    assert os.path.exists(args.perforce_content_dir)
+
+    unreal_project_content_dir = os.path.realpath(os.path.join(os.path.dirname(__file__), "..", "cpp", "unreal_projects", "SpearSim", "Content"))
+
+    # Create and/or remove symlink for Shared directory
+    perforce_content_shared_dir = os.path.realpath(os.path.join(args.perforce_content_dir, "Shared"))
+
```
RachithP, line 24: To maintain consistency, it'd be good to 
```python
assert os.path.exists(perforce_content_shared_dir)
```
similar to 
```python
assert os.path.exists(perforce_content_scene_dir)
``` 
below.

mikeroberts3000: Fair point. Only one of them depends on an additional input argument that hasn't already been validated, so strictly speaking, only `perforce_content_scene_dir` really needs to be checked. But there is no harm in asserting for both, and I like the added consistency.

tools/prepare_scene_for_editing.py
```diff
@@ -0,0 +1,54 @@
...
+    # We do not want to use os.path.realpath(...) here, because that will resolve to the directory inside the user's Perforce workspace.
+    # Instead, we want this path to refer to the symlinked version inside the user's unreal project directory.
+    unreal_project_content_shared_dir = os.path.join(unreal_project_content_dir, "Shared")
+
+    if spear.path_exists(unreal_project_content_shared_dir):
+        print(f"[SPEAR | prepare_scene_for_editing.py] File or directory or symlink exists, removing: {unreal_project_content_shared_dir}")
+        spear.remove_path(unreal_project_content_shared_dir)
+
+    if not args.remove_symlink:
```
RachithP, line 34: I don't think we need the `remove_symlink` argument or rather I don't understand it's purpose.
- If the user does not pass `--remove_symlink`, then we are creating a symlink. (This makes sense and should be the default which it is currently)
- If the user passes `--remove_symlink`, then we are not doing anything here. In this case, why would we run this script?

mikeroberts3000: Conceptually, the script has two modes: "create" and "remove". I want the create mode because sometimes I want to be able to edit a given scene in the editor. I want the remove mode because I typically don't want to cook an executable when I've got extra scenes in my `Content` folder (e.g., this will slow down the cooking process, there might be cooking errors, etc), so I want a way to remove those scenes from my content folder.

As an implementation detail, the create mode also happens to remove an existing symlink because that is the safest and most robust thing to do. Creating a symlink on top of an existing file/folder/symlink-from-another-platform/etc can cause problems. In this sense, the functionality in the create mode is a strict superset of the functionality in the remove mode.

Create mode is considered the default behavior in this script, and remove mode is activated by the `--remove_symlink` flag. We're not doing nothing if the user passes in `--remove_symlink`. We're removing the existing symlink and not creating a new one.

RachithP (approved):
I have few comments on `prepare_scene_for_editing.py`.

Rest Looks good!

tools/prepare_scene_for_editing.py
mikeroberts3000: But we're creating one in it's place on the very next line of code. I don't think it's misleading. `--remove_symlink` refers to the logical outcome of the entire script. If you don't pass in `--remove_symlink`, the logical outcome is that new symlinks will be created. You can think of it as existing symlinks getting updated if you prefer. It just so happens that the update operation is implemented in terms of a remove+add operation.

PR #169, opened 2023-03-24
Open Loop Control for Fetch in Kujiale Scene

rwang15:

PR #176, opened 2023-03-28
specify the spawn location of actors

mikeroberts3000:
- the user can specify the spawn location of actors, which is useful when the task is set to NullTask
- this PR is cherry-picked and lightly modified from https://github.com/isl-org/spear/pull/169, so we can start building public release executables before that PR is finalized


python/spear/config/default_config.simulation_controller.yaml
```diff
@@ -20,11 +20,19 @@ SIMULATION_CONTROLLER:
   CUSTOM_UNREAL_CONSOLE_COMMANDS: []
 
   CAMERA_AGENT:
+    CAMERA_ACTOR_NAME: ""
     ACTION_COMPONENTS: ["set_pose", "set_num_random_points"] # "set_pose", "set_num_random_points"
     OBSERVATION_COMPONENTS: ["camera"] # "camera"
     STEP_INFO_COMPONENTS: ["random_points"] # "random_points"
+    SPAWN_MODE: "specify_pose" # "specify_pose", "specify_existing_actor"
```
RachithP, line 27: Would it be better to keep the default as empty string? This way, user has to define a certain spawn mode, and combining this with other consequent parameters like spawn position, orientation, etc is more clear in config yamls. In our `user_config.yaml.example` file, we can keep the default as `"specify_poses"`.

RachithP (approved):
Looks good!

python/spear/config/default_config.simulation_controller.yaml
mikeroberts3000: I like `specify_pose` as the default, because I think spawning an actor at `[0,0,0]` is a reasonable default (this is the existing behavior), and that is what you'll get if you don't specify anything extra in `user_config.yaml`. Also, actors will sometimes be re-positioned by the task anyway, so I wouldn't want to force those users to specify more stuff than necessary in their `user_config.yaml`.

PR #177, opened 2023-03-28
Fixes and improvements

RachithP:
- Fix issue on Linux when using `run_executable.py`. Need to set `VK_ICD_FILENAMES` env variable before launching the script.
- `generate_images.py` - Improved directory structure for storing images when multiple scenes are present.

### DIFF
--- a/examples/generate_image_dataset/generate_images.py
+++ b/examples/generate_image_dataset/generate_images.py
@@ -107,7 +107,7 @@ if __name__ == "__main__":
             # create dir for storing images
             if not args.benchmark:
                 for render_pass in config.SIMULATION_CONTROLLER.CAMERA_AGENT.CAMERA.RENDER_PASSES:
-                    render_pass_dir = os.path.realpath(os.path.join(args.images_dir, render_pass))
+                    render_pass_dir = os.path.realpath(os.path.join(args.images_dir, pose["scene_id"], render_pass))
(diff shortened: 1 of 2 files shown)


examples/generate_image_dataset/generate_images.py
```diff
@@ -107,7 +107,7 @@ def single_step(self, action=None, get_observation=False):
             # create dir for storing images
             if not args.benchmark:
                 for render_pass in config.SIMULATION_CONTROLLER.CAMERA_AGENT.CAMERA.RENDER_PASSES:
-                    render_pass_dir = os.path.realpath(os.path.join(args.images_dir, render_pass))
+                    render_pass_dir = os.path.realpath(os.path.join(args.images_dir, pose["scene_id"], render_pass))
```
mikeroberts3000, line 110: I don't think this added level of hierarchy is necessary. We're using a global index for the filename that corresponds directly to the row in poses.csv, so I think it's slightly cleaner to have all the images in a single folder.

tools/run_executable.py
```diff
@@ -17,13 +17,19 @@
     parser.add_argument("--paks_dir")
     parser.add_argument("--scene_id")
     parser.add_argument("--map_id")
+    parser.add_argument("--vulkan_device_files")
```
mikeroberts3000, line 20: Great catch.

mikeroberts3000 (approved):

PR #178, opened 2023-03-30
Fixes + improvements

RachithP:
* Fix low FPS issue on warehouse_0000 scene.
* Fix typo in run.py and some improvements.
* Add comment to the issue with GenerateProjectFiles.bat/sh not working on Linux, Windows.


python/spear/config/default_config.simulation_controller.yaml
```diff
@@ -15,6 +15,14 @@ SIMULATION_CONTROLLER:
   PORT: 30000
 
   SIMULATION_STEP_TIME_SECONDS: 0.1
+  ENABLE_ENHANCED_DETERMINISM: True
+  ENABLE_SUBSTEPPING: True
+  MAX_SUBSTEP_DELTA_TIME: 0.01
+  MAX_SUBSTEPS: 100
+  ContactOffsetMultiplier: 0.010000
```
mikeroberts3000, line 22: CONTACT_OFFSET_MULTIPLIER (we should stick to our all-caps naming convention)

mikeroberts3000: move this section of parameters to be after CUSTOM_UNREAL_CONSOLE_COMMANDS but before CAMERA_AGENT

mikeroberts3000: no need for enhanced determinism and substepping because updating them here has no effect

mikeroberts3000: no need to say 0.0100000, just say 0.01

examples/open_loop_control_fetch/user_config.yaml.example
```diff
@@ -15,6 +15,8 @@ SPEAR:
   WINDOW_RESOLUTION_Y: 512
 
 SIMULATION_CONTROLLER:
+  MAX_SUBSTEP_DELTA_TIME: 0.001
```
mikeroberts3000, line 18: move after CUSTOM_UNREAL_CONSOLE_COMMANDS so it's in the same order as in default_config

examples/open_loop_control_fetch/run.py
```diff
@@ -75,11 +74,6 @@
 
     if args.benchmark:
         start_time_seconds = time.time()
-    else:
```
mikeroberts3000, line 78: - leave this else statement here, but include a for-loop over rendering passes, for consistency with generate_images.py (line 109)
- this is slightly cleaner than doing the operation repeatedly, because conceptually this should be a one-time operation

examples/imitation_learning_openbot/user_config.yaml.example
```diff
@@ -27,6 +27,8 @@ SPEAR:
   WINDOW_RESOLUTION_Y: 120
 
 SIMULATION_CONTROLLER:
+  MAX_SUBSTEP_DELTA_TIME: 0.001
```
mikeroberts3000, line 30: move after CUSTOM_UNREAL_CONSOLE_COMMANDS

python/spear/config/default_config.simulation_controller.yaml
mikeroberts3000: in default_config, set {ContactOffsetMultiplier, MinContactOffset, MaxContactOffset} to their default values in the Unreal Engine, and then set them to these weird values per-example on an as-needed basis (do we need these values for the OpenBot example?)

cpp/unreal_projects/SpearSim/Source/SpearSim.Target.cs
```diff
@@ -44,9 +42,11 @@ public SpearSimTarget(TargetInfo Target) : base(Target)
                     Path.Combine(plugin, "ThirdParty") + "=" +
                     Path.GetFullPath(Path.Combine(ProjectFile.Directory.FullName, "..", "..", "..", "third_party"));
             }
-            
-        } else {
-            throw new Exception("[SPEAR | SpearSim.Target.cs] Target.Platform == " + Target.Platform);            
+        // we can't throw an exception here because when we invoke GenerateProjectFiles.bat/GenerateProjectFiles.sh script, Unreal Build Tool
```
mikeroberts3000, line 45: Add whitespace before this comment, and move it outside of the the else-if statement
```

        // We can't throw an exception here, because when we invoke GenerateProjectFiles.bat or GenerateProjectFiles.sh (packaged with
        // the Unreal Engine), instances of this class get created with Target.Platform set to many different platforms (e.g., iOS, tvOS,
        // Win32, etc).
        // } else {
        //    throw new Exception("[SPEAR | SpearSim.Target.cs] Target.Platform == " + Target.Platform);            
        // }
```

mikeroberts3000 (approved):
Several comments to address but otherwise good.

examples/open_loop_control_fetch/run.py
RachithP: I don't think we are doing `shutil.rmtree(args.image_dir, ignore_errors=True) in our generate image dataset example, which is why I removed it from here. Let me add it our generate image dataset example as well.

python/spear/config/default_config.simulation_controller.yaml
RachithP: > no need for ENABLE_ENHANCED_DETERMINISM and ENABLE_SUBSTEPPING because we found that updating them in the yaml has no effect

They do have an effect when updated via yaml file. It's just that the values need to be set initially in config file for the current set of actions in the urdf open loop control example to work.

RachithP: > in default_config, set {ContactOffsetMultiplier, MinContactOffset, MaxContactOffset} to their default values in the Unreal Engine, and then set them to these weird values per-example on an as-needed basis (do we need the weird values for the OpenBot example?)

I'm not sure if we need these for OpenBot example. Not setting these weird values did not seem to affect the way OpenBot navigated in the environment.

PR #185, opened 2023-04-17
Mention UE 4.26 paths in doc

samarth-robo:
TODO: paths for Windows and Linux.


mikeroberts3000:
I updated the docs to include this change and various other details, including the default paths for other Windows and Linux.

PR #194, opened 2023-04-22
fixes needed for MuJoCo to correctly model Fetch

samarth-robo:
To reproduce the error in MuJoCo:
- Unzip [urdf_meshes.zip](https://github.com/isl-org/spear/files/11300003/urdf_meshes.zip) containing STL meshes in to a directory called `urdf_meshes`.
- Run the following code (`convert_urdf.py`) to get an updated URDF with paths pointing to the STL meshes in `urdf_meshes`: `$ python convert_urdf.py --input fetch.xml --meshdir urdf_meshes`:
```python
import argparse
import xml.etree.ElementTree as ET
import os


osp = os.path


if __name__ == '__main__':
  parser.add_argument('--input', required=True, help='Input URDF file')
  parser.add_argument('--meshdir', required=True, help='Directory containing ')
  args = parser.parse_args()
  
  filename = osp.expanduser(args.input)
  tree = ET.parse(filename)
  root = tree.getroot()

  # add mujoco tag and compiler options
  mujoco = ET.SubElement(root, 'mujoco')
  compiler = {'meshdir': osp.expanduser(args.meshdir), 'discardvisual': 'false'}
  compiler = ET.SubElement(mujoco, 'compiler', compiler)

  for child in root:
    if child.tag == 'link':
      visual = child.find('visual')
      visual_mesh = visual.find('geometry').find('mesh')
      collision = child.find('collision')
      collision_mesh = collision.find('geometry').find('mesh')
      if visual_mesh is None:
        visual_name = f'{child.get("name")}_visual'
        collision_name = f'{child.get("name")}_collision'
      else:
        asset_name = visual_mesh.get('filename')
        mesh_filename = asset_name[12:-1]
        mesh_filename = osp.split(mesh_filename)[-1].split('.')[-1]
        visual_name = f'{mesh_filename}_visual'
        visual_mesh.set('filename', f'{mesh_filename}_visual.stl')
        collision_name = f'{mesh_filename}_collision'
        collision_mesh.set('filename', f'{mesh_filename}_collision.stl')
      visual.set('name', visual_name)
      collision.set('name', collision_name)
      ET.SubElement(child, 'spear', {'assetpath': asset_name})

  out_filename = 'output.xml'
  tree.write(out_filename)
  print(f'{out_filename} written')
```
- It will produce `output.xml`. In a Python env containing installed deepmind/mujoco, run `$ python -m mujoco.viewer --mjcf output.xml`.

### DIFF
--- a/python/spear/urdf/fetch.xml
+++ b/python/spear/urdf/fetch.xml
@@ -522,20 +522,20 @@
             <inertia ixx="6.98014e-06" ixy="-2.62572e-08" ixz="-1.58206e-08" iyy="4.12665e-05" iyz="-1.57909e-09" izz="3.75763e-05"/>
         </inertial>
         <visual>
-            <origin rpy="0 0 0" xyz="0 -0.101425 0"/>
+            <origin rpy="0 0 0" xyz="0 0.101425 0"/>
             <geometry>
                 <mesh filename="StaticMesh'/UrdfBot/Fetch_V1/Meshes/gripper_finger_link_l.gripper_finger_link_l'"/>
             </geometry>
         </visual>
         <collision>
-            <origin rpy="0 0 0" xyz="0 -0.101425 0"/>
+            <origin rpy="0 0 0" xyz="0 0.101425 0"/>
             <geometry>
                 <mesh filename="StaticMesh'/UrdfBot/Fetch_V1/Meshes/gripper_finger_link_l.gripper_finger_link_l'"/>
             </geometry>
         </collision>
     </link>
     <joint name="gripper_finger_joint_l" type="prismatic">
-        <origin rpy="0 0 0" xyz="0 -0.015425 0"/>
+        <origin rpy="0 0 0" xyz="0 0 0"/>
         <parent link="gripper_link"/>
         <child link="gripper_finger_link_l"/>
         <axis xyz="0 -1 0"/>
@@ -550,20 +550,20 @@
             <inertia ixx="6.15416e-06" ixy="-1.5025e-09" ixz="0" iyy="3.87474e-05" iyz="0" izz="3.55162e-05"/>
         </inertial>
         <visual>
-            <origin rpy="0 0 0" xyz="0 0.101425 0"/>
+            <origin rpy="0 0 0" xyz="0 -0.101425 0"/>
             <geometry>
                 <mesh filename="StaticMesh'/UrdfBot/Fetch_V1/Meshes/gripper_finger_link_r.gripper_finger_link_r'"/>
             </geometry>
         </visual>
         <collision>
-            <origin rpy="0 0 0" xyz="0 0.101425 0"/>
+            <origin rpy="0 0 0" xyz="0 -0.101425 0"/>
             <geometry>
                 <mesh filename="StaticMesh'/UrdfBot/Fetch_V1/Meshes/gripper_finger_link_r.gripper_finger_link_r'"/>
             </geometry>
         </collision>
     </link>
     <joint name="gripper_finger_joint_r" type="prismatic">
-        <origin rpy="0 0 0" xyz="0 0.015425 0"/>
+        <origin rpy="0 0 0" xyz="0 0 0"/>
         <parent link="gripper_link"/>
         <child link="gripper_finger_link_r"/>
         <axis xyz="0 1 0"/>
--- a/cpp/unreal_plugins/UrdfBot/Source/UrdfBot/UrdfLinkComponent.cpp
+++ b/cpp/unreal_plugins/UrdfBot/Source/UrdfBot/UrdfLinkComponent.cpp
@@ -39,7 +39,7 @@ void UUrdfLinkComponent::initializeComponent(UrdfLinkDesc* link_desc)
         float m_to_cm = 100.0f;
         UrdfLinkDesc* parent_link_desc = joint_desc->parent_link_desc_;
         FTransform link_origin_ = parent_link_desc->visual_descs_[0].origin_;
-        SetRelativeLocation((joint_desc->origin_.GetLocation() + link_desc->visual_descs_[0].origin_.GetLocation() - link_origin_.GetLocation()) * m_to_cm);
+        SetRelativeLocation((joint_desc->origin_.GetLocation() - link_desc->visual_descs_[0].origin_.GetLocation() - link_origin_.GetLocation()) * m_to_cm);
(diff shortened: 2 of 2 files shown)


python/spear/urdf/fetch.xml
```diff
@@ -522,20 +522,20 @@
             <inertia ixx="6.98014e-06" ixy="-2.62572e-08" ixz="-1.58206e-08" iyy="4.12665e-05" iyz="-1.57909e-09" izz="3.75763e-05"/>
         </inertial>
         <visual>
-            <origin rpy="0 0 0" xyz="0 -0.101425 0"/>
+            <origin rpy="0 0 0" xyz="0 0.101425 0"/>
```
mikeroberts3000, line 525: Do these changes need to be accompanied by corresponding changes to our `UrdfBot` plugin, so that our `open_loop_control_fetch` demo still behaves as expected?

mikeroberts3000:
Is the output of `convert_urdf.py` a fixed URDF file (e.g., with fixed mesh filenames)? Or is it an MJCF fragment that can be included from another MJCF file? I don't know exactly what `--mjcf` does when invoking the Mujoco viewer. (You'd think it would mean specifying an MJCF file, but I don't see us creating a `worldbody` node anywhere when we generate `output.xml`, so I'm not sure.)

- If it is the former (fixed URDF file), it is ok not to check in `convert_urdf.py`. We can leave it as a comment in this PR, or if you prefer, you can check it into our `spear-experimental` repo in the `experimental` directory.
- If it is the latter (MJCF fragment), let's add `convert_urdf.py` to the `tools` folder but with the following changes:
  - Let's call it `generate_mjcf.py` because our convention for naming `tools` is to use a verb-noun pair, where the noun is the output (e.g., MJCF) rather than the input (e.g., URDF).
  - Let's separate the URDF fixup logic from the conversion-into-MJCF logic. `generate_mjcf.py` should assume well-formed URDF input, e.g., passing through the mesh filenames exactly as they are specified in the input URDF file.

python/spear/urdf/fetch.xml
```diff
@@ -522,20 +522,20 @@
...
             <geometry>
                 <mesh filename="StaticMesh'/UrdfBot/Fetch_V1/Meshes/gripper_finger_link_l.gripper_finger_link_l'"/>
             </geometry>
         </visual>
         <collision>
-            <origin rpy="0 0 0" xyz="0 -0.101425 0"/>
+            <origin rpy="0 0 0" xyz="0 0.101425 0"/>
             <geometry>
                 <mesh filename="StaticMesh'/UrdfBot/Fetch_V1/Meshes/gripper_finger_link_l.gripper_finger_link_l'"/>
```
mikeroberts3000, line 533: Since we're changing the URDF file anyway, we should also change it to use our custom `spear` tags where appropriate, and make the corresponding changes to the `UrdfBot` plugin at the same time. Since we will be maintaining URDF files and MJCF files going forward, let's name this file `fetch.urdf` instead of `fetch.xml` to avoid confusion.

samarth-robo:
> Is the output of `convert_urdf.py` a modified URDF file (e.g., with modified mesh filenames)? Or is it an MJCF fragment that can be included from another MJCF file? I don't know exactly what `--mjcf` does when invoking the Mujoco viewer. (You'd think it would mean specifying an MJCF file, but I don't see us creating a `worldbody` node anywhere when we generate `output.xml`, so I'm not sure.)
> 
> * If it is the former (modified URDF file), it is ok not to check in `convert_urdf.py`. We can leave it as a comment in this PR, or if you prefer, you can check it into our `spear-experimental` repo in the `experimental` directory.
> * If it is the latter (MJCF fragment), let's add `convert_urdf.py` to the `tools` folder but with the following changes:
>   
>   * Let's call it `generate_mjcf.py` because our convention for naming `tools` is to use a verb-noun pair, where the noun is the output (e.g., MJCF) rather than the input (e.g., URDF).
>   * Let's separate the URDF fixup logic from the conversion-into-MJCF logic. `generate_mjcf.py` should assume well-formed URDF input, e.g., passing through the mesh filenames exactly as they are specified in the input URDF file.

It generates a modified URDF file. I agree it is not a well-named command-line argument, but you can pass a URDF to `--mjcf`.

python/spear/urdf/fetch.xml
samarth-robo: For links, it `convert_urdf.py` already generates a `<spear>` subtag with the `assetpath` attribute pointing to the .uasset file. I think we can keep the URDF joints untouched, because we have to add actuators to the MJCF separately anyway.
Agree with the renaming.

python/spear/urdf/fetch.xml
mikeroberts3000: Let's assume: (1) the URDF-to-MJCF parsing code in this PR is correct; (2) Mujoco loads `output.xml` correctly; and (3) all the meshes in `urdf_meshes.zip` have been exported correctly.

Under these assumptions, we can use Mujoco's behavior when loading MJCF files as a reference for how SPEAR should behave. In other words, the Fetch robot spawned by the `UrdfBot` plugin in SPEAR should match the Fetch robot spawned in Mujoco. Ideally, we would match with Mujoco numerically and visually, but if we need to choose one or the other, we should match numerically.

If SPEAR does not match the behavior of Mujoco, we should fix our `UrdfBot` code in this PR, and also apply the coordinate system change above, because then we'll have exactly the same URDF behaving exactly the same way in SPEAR and Mujoco.

python/spear/urdf/fetch.xml
mikeroberts3000: Let's check in the updated `fetch.xml` in this PR with the following changes:
- rename `fetch.xml` to `fetch.urdf` to avoid confusion with MJCF that we will generate later
- rename `assetpath` to `unreal_asset`
- fix the indentation (when I ran the code above, the `spear` element was not indented correctly)

mikeroberts3000:
> It generates a modified URDF file. I agree it is not a well-named command-line argument, but you can pass a URDF to `--mjcf`.

That makes sense.

mikeroberts3000:
We should also add a custom `spear` element inside the `material` element, because this is another place where we're overloading/abusing an existing attribute to store custom data, e.g.,

```
    <material name="caster_link_material">
        <texture filename="" />
        <spear unreal_asset="/UrdfBot/Common/M_Slippery.M_Slippery" />
    </material>
```

PR #196, opened 2023-04-27
fix UE 5.2 build issue related to symbolic links

mikeroberts3000:
- fixes UE 5.2 build issue related to symbolic links
- safer handling of symbolic links in path remapping code (depending on how the -ffile-prefix-map argument is handled in clang, our previous code would potentially prevent stepping through third-party code, but this PR contains a fix)


RachithP (approved):
Looks good!

PR #200, opened 2023-05-13
Scene Data Structure Example

rwang15:
Note that it depend on Shared/kujiale folder on Perforce (blocked by gitignore).


mikeroberts3000:
- What are the `SpearSim/Content/Shared/kujiale/Textures/*.uasset` textures used for?
- Do we need `SpearSim/Content/Shared/Koolab` and `SpearSim/Content/Shared/kujiale`, or will one eventually replace the other?
- Move `SpearSim/Content/Shared/kujiale` to `SpearSim/Content/Kujiale`, use `M_` prefix for material files, `T_` for textures, etc.

mikeroberts3000:
- Do the folders in `SpearSim/Content/Scenes/debug_0001/Materials` need to have such opaque names? It's fine if this is absolutely necessary, but are there any other options to assign more human-readable names here? Even if these opaque names are necessary, they are inconsistent, e.g., `5ac9e5d43cf8ab2e12006b69` and `PureColorMaterial--6710887`.
- The individual assets in `SpearSim/Content/Scenes/debug_0001/Materials/*` are also inconsistently named, e.g., `M_M_M_6276472ca7755a0001f30538` and `MI_54213fb3254e415213ef0ffc_index17`.

mikeroberts3000:
**SpearSim/Content/Scenes/debug_0001/Meshes**

- It's hard to browse `SpearSim/Content/Scenes/debug_0001/Meshes/*`. What is the difference between `Architecture` and `Furniture`? Does "architecture" mean all the stuff that would already be there when you move into a brand new apartment (e.g., floors, walls, ceilings), and "furniture" mean all the clutter that you would bring to the apartment (e.g., tables, chairs)? I don't think "architecture" and "furniture" are the right words to capture this distinction. "Architecture" doesn't convey the intended meaning, and there are many object categories that wouldn't be considered furniture in the "furniture" folder. I have proposed an alternative convention in our SPEAR v0.3.0 design doc.

mikeroberts3000:
**World Outliner**

- I have proposed an alternative convention for arranging our project in the World Outliner in our SPEAR v0.3.0 design doc.

mikeroberts3000:
I updated the SPEAR v0.3.0 design doc (and my comments above) based on our most recent technical discussion. There are still some refinements we should make, but I think we're getting close.

mikeroberts3000:
- The way outdoor lighting is set up in this scene is also inconsistent with how we represent outdoor illumination in `debug_0000`. I'm not sure which setup is better. Until we have a compelling reason to choose one setup or the other, I'd rather do it the `debug_0000` way because that is how outdoor lighting is configured in the starter content.

mikeroberts3000:
Comments on `debug_0001`:

- I don't think `door_00.physics_constraint_00` is set up correctly. I attempted to move `door_00` into free space and rotate it slightly so that the door will swing open due to gravity. However, when I press play, the door doesn't swing open. I think we should set up the constraint in an analogous way to the constraints in `debug_0000`. The `ahlam/apartment_0000_scene` branch also has a bunch of correctly functioning drawers and cupboards to use as a reference.
- I suspect it will be better for downstream applications (e.g., exporting to Mujoco, exporting to URDF) if the constraint here was between a parent and child, rather than between two siblings. The URDF format only supports kinematic trees, so I don't think it is possible to describe `door_00.physics_constraint_00` in a URDF file. On the other hand, Mujoco has special facilities to model additional constraints between siblings (https://mujoco.readthedocs.io/en/stable/overview.html#stand-alone-elements), but based on the documentation, these facilities seem inefficient, and it seems like they won't take advantage of Mujoco's efficient joint-space dynamics formulation.

mikeroberts3000:
Comments on debug_0001:

- `ceiling_cover_00` has collisions turned off, which shows up as a non-default choice in the Details pane. Is this intentional?

mikeroberts3000:
Comments on `kujiale_0000`:

- Most of my comments on `debug_0000` above also apply to `kujiale_0000`, so I won't attempt to repeat them.
- The many distinct ceiling static meshes should be grouped under a single actor so the default behavior is to move them as a group. Likewise for wall meshes and floor meshes.
- It would be even better if we could group the floor meshes into a one-actor-per-room grouping, because that would essentially provide a room-level segmentation for free. (Is the name of each room available in any of the Kujiale metadata?) But if this is not possible, then it would be better to group the floor meshes into a single logical actor. There are currently many more floor actors than there are rooms. It would be better if these meshes could all be moved in more logical entities.
- Why are the floor meshes stored in the `Base/Unknown` folder instead of the `Base/Floor` folder?
- The `floor_40` object is actually on the ceiling.
- There are two objects named `floor_00`.
- There are two `door_handle` objects, but they're touching a `window`, not a `door`.

mikeroberts3000:
Comments on `kujiale_0000`:

- I can get back up to 30fps when I disable all the `08_chandelier`, `09_ceiling_light`, `10_wall_light`, `12_table_lamp`, `19_cabinet` meshes, as well as the `Light Source` directional light, but I don't understand why Unreal would be having trouble with this combination of objects. (Enabling and disabling different categories of objects like this is an early win for our new project structure. I would have no idea how to enable different different categories of objects like this in our old project structure.)
- Possible issues:
  - The `Nanite Visualization -> Overview` visualization shows a surprisingly high amount of overdraw, i.e., Nanite is trying to render to each pixel many times. According to this Nanite talk (https://youtu.be/TMorJX3Nj6U?t=2386), overdraw can happen when there are two meshes that are very close together.
  - The `stat gpu` view shows that the "Basepass" gets significantly more expensive when the light sources are enabled. I don't think this happens for `debug_0000` and `apartment_0000`. What is unique about `kujiale_0000`, and why are the cabinets a performance hot spot?

mikeroberts3000:
Comments on `kujiale_0000`:

- The names of static meshes can be left as-is in the Content Browser, because we have already given up on human-readability in the Content Browser. But it would be better to have a more consistent naming convention for actors with nested static mesh components. For example, `door_handle_00` has a static mesh component called `obj0`. It would be better to call these `mesh_00`, `mesh_01`, etc to match the other composite objects in the scene.
- There are multiple `17_shelf` objects that could be grouped into a single logical entity. Some `17_shelf` objects are actually coat hangers, not shelves. `shelf_03` is floating in space.

mikeroberts3000:
Comments on `kujiale_0000`:

- The navmesh is missing some non-trivial floor regions.

mikeroberts3000:
Update: the performance issues with `kujiale_0000` I described above are the roughly the same after upgrading from a 2080 to a 4090 with up-to-date drivers.

mikeroberts3000:
Comments on `kujiale_0000`:

- The collision meshes for `wall_40` and `wall_44` are mis-shaped. I suspect an underlying cause might be that the rendering meshes for these objects are L-shaped, whereas the other rendering meshes for walls in this scene are flat. This would cause the convex approximation of the L-shaped object to diverge significantly from its rendering mesh.

mikeroberts3000:
Comments on `kujiale_0000`:

Only one of these materials is referenced in the scene according to `Asset Actions > Select Actors Using This Asset`. Why don't these materials match everything else in the `Materials` folder?

```
/Script/Engine.MaterialInstanceConstant'/Game/Scenes/kujiale_0000/Materials/5642e468d3e4d2d8a6406bad/MI_5642e468d3e4d2d8a6406bad.MI_5642e468d3e4d2d8a6406bad'
/Script/Engine.MaterialInstanceConstant'/Game/Scenes/kujiale_0000/Materials/5642e468d3e4d2d8a6406badC/MI_5642e468d3e4d2d8a6406badC.MI_5642e468d3e4d2d8a6406badC'
/Script/Engine.MaterialInstanceConstant'/Game/Scenes/kujiale_0000/Materials/5642e468d3e4d2d8a6406badCC/MI_5642e468d3e4d2d8a6406badCC.MI_5642e468d3e4d2d8a6406badCC'
/Script/Engine.MaterialInstanceConstant'/Game/Scenes/kujiale_0000/Materials/5642e468d3e4d2d8a6406badCCC/MI_5642e468d3e4d2d8a6406badCCC.MI_5642e468d3e4d2d8a6406badCCC'
/Script/Engine.MaterialInstanceConstant'/Game/Scenes/kujiale_0000/Materials/5642e468d3e4d2d8a6406badCCCC/MI_5642e468d3e4d2d8a6406badCCCC.MI_5642e468d3e4d2d8a6406badCCCC'
/Script/Engine.MaterialInstanceConstant'/Game/Scenes/kujiale_0000/Materials/5642e468d3e4d2d8a6406badCCCCC/MI_5642e468d3e4d2d8a6406badCCCCC.MI_5642e468d3e4d2d8a6406badCCCCC'
```

mikeroberts3000:
Can you please pull the latest changes from `ue-52` into this branch? It will make it easier to toggle back and forth between branches, which in turn will make it easier for us to review things together. Currently there are mismatches with the `third_party` folder because this branch is somewhat out-of-date. Since this branch only adds to the Content folder, and doesn't modify existing code, it should be relatively straightforward.

rwang15:
A few TODO items: 
1. The navmesh is missing some non-trivial floor regions
	1. SM_Dummy are take into account during navmesh computation, and they do not lie along with actual static mesh.  Maybe get rid of the SM_Dummmy and use its first child staticmesh?
2. The collision meshes for wall_40 and wall_44 are mis-shaped
	1. will try to see if using multiple component structure as wall works for light leaking
3. unused Materials in Materials folder
	1. they are originally used in the scene but replaced by other materials from Megascan. 
4. StaticMeshes in Base folder does not match their semantic label
	1. they are not labeled correctly at first. I manually changed their semantic label in the world editor but not in the Content Folder. 
5. More than 100 actor in a single semantic category. 
	1. NO

mikeroberts3000:
One-time error when opening `debug_0001`:

```
Map check complete: 0 Error(s), 2 Warning(s), took 0.088ms to complete.
M_TableRound Material /Game/StarterContent/Props/Materials/M_TableRound.M_TableRound was missing the usage flag bUsedWithNanite. If the material asset is not re-saved, it may not render correctly when run outside the editor. Fix
M_Chair Material /Game/StarterContent/Props/Materials/M_Chair.M_Chair was missing the usage flag bUsedWithNanite. If the material asset is not re-saved, it may not render correctly when run outside the editor. Fix
```

mikeroberts3000:
- Use consistent number of digits for `_obj6` and similar suffixes, e.g., `SM_Wall_0075` and `SM_Clutter_2345` and `SM_Clutter_2345_Part_1341`.
- We need a consistent rule for setting "Collision Presets" on static meshes. Here is a suggested convention. Set SM_Dummy to have the following collision properties in the static mesh editor. "Collision Enabled" should be set to "Probe Only", because we want to enable the "Simulate Physics" option in the Details Pane for our dummy components, so we can't disable collisions entirely. Setting "Collision Enabled" to "Probe Only" will ensure that our dummy components don't actually generate collisions, but it will also also allow us to enable the "Simulate Physics" option without generating any warnings. Set "Object Type" to "WorldStatic" because that is what the other meshes in StarterContent are set to. Now we can set all StaticMeshComponents in the Details Pane to use the "Default" collision preset if they have a static mesh assigned to them, or "BlockAllDynamic" if they do not. If a StaticMeshComponent does not have a static mesh assigned to it, the "Default" collision preset is not available, and "BlockAllDynamic" is the default value. This is the convention in debug_0000.
- Rename `ceiling_slab -> ceiling_cover_00`, `floor_slab -> floor_cover_00` for consistency with `kujiale_0000` and use numbers because there might be multiple of them.
- `ceiling_00` does not appear to be visible in the scene from above or below. I think it's at the wrong location.
- `Sphere` is located outside the room. If this object is useful for debugging, let's keep it inside the room and put it in the `10_other` folder with a name that is consistent with the other actors. If it is not useful for debugging, let's remove it.

mikeroberts3000:
Assets are renamed but still are not using a consistent number of digits. These assets won't sort correctly in Python. Example: `SM_Floor_0` uses 1 digit but `SM_Door_19` uses 2 digits.

mikeroberts3000:
- Use 4 digits for all clutter meshes, e.g., `SM_Clutter_000`, for consistency.
- Use 4 digits for the part IDs too, e.g., `SM_Clutter_0001_part_0000`.
- Why does `Clutter/MKTQX...` start with 1-indexing (i.e., `SM_Clutter_0001`) but all the other clutter folders start with 0-indexing (i.e., `SM_Clutter_0000`)?

mikeroberts3000 (approved):
Please address my final comments, but otherwise this PR looks great!

rwang15:
MKTQXMYKTEXM6AABAAAAAAA8 comes from a different pipeline and I have manually rename it to match current format. All other assets are from the same pipeline and should follow the same naming convension.

RachithP:
@rwang15 @mikeroberts3000 Now that we have `cpp/unreal_projects/SpearSim/Content/Kujiale` folder contents as part of this repo, do we need the `Kujiale` folder on Perforce? From comparing the two folders, they appear to contain the same/similar assets.

If we do not need it, the following tasks need to be completed to ensure consistency with this PR:
- Remove `Kujiale` folder from Perforce
- Update `build_paks.py`  (this seems to be already updated)
- Update `prepare_scene_for_editing.py`

PR #202, opened 2023-05-14
logging and other improvements

mikeroberts3000:
- improved printing and logging functionality in C++/C#/Python
- improved CameraSensor to return images that match the game viewport pixel-perfectly
- refactored common functionality in Build.cs and Target.cs files to avoid code duplication
- implemented a lightweight and extendable GUI inside the Unreal Engine
- implemented a system for intercepting console commands inside and outside of the editor, included some example commands
- implemented simple physics in the debug_0000 scene as an example of how to structure composite objects (i.e., multiple meshes) that are articulated/dynamic/static and composite light sources (i.e., emissive mesh and light)
- fixed bug with creating symlinks to the Megascans and MSPresets shared folders in Perforce
- cleaned up include statements and applied include-what-you-use rule more strictly throughout the code
- renamed Box to ArrayDesc for clarity and consistency
- renamed RenderPass to RenderPassDesc for clarity and consistency
- renamed UActorHitEvent to UActorHitEventComponent for clarity and consistency
- renamed UTickEvent to UTickEventComponent for clarity and consistency
- renamed macros to have SP_ prefix for consistency with SP_LOG
- removed GPULightMass plugin because we're not baking lighting any more
- miscellaneous refactoring and cleanup


mikeroberts3000:
Note that I did not attempt to integrate any of these changes into OpenBot or UrdfBot, since those plugins are being actively developed and I didn't want to create major conflicts. The old ASSERT and SUPPRESS_COMPILER_WARNINGS macros are still available in this PR, so we can migrate to the new macros incrementally.

cpp/unreal_plugins/CoreUtils/Source/CoreUtils/Log.h
```diff
@@ -0,0 +1,52 @@
+//
+// Copyright(c) 2022 Intel. Licensed under the MIT License <http://opensource.org/licenses/MIT>.
+//
+
+#pragma once
+
+#include <string>
+
+#include <filesystem>
+#include <utility>
```
RachithP, line 10: Combine these and sort alphabetically since they all are part of standard library?

mikeroberts3000: Ah yes of course. Good catch. I used to be including <string.h> which I considered to be part of the lower-level C standard library, but now that I'm using <string> it should be alphabetical with the rest of the C++ standard library.

tools/build_paks.py
```diff
@@ -110,16 +117,16 @@
...
             os.symlink(perforce_content_scene_dir, unreal_project_content_scene_dir)
 
-        # Now that we have created a symlink, our Unreal project should contain exactly two scenes: starter_content_0000 and scene_id
+        # Now that we have created a symlink, our Unreal project should contain exactly two scenes: debug_0000 and scene_id
         ignore_names = [".DS_Store"]
         unreal_project_scenes = { os.path.basename(x) for x in os.listdir(unreal_project_content_scenes_dir) if x not in ignore_names }
-        assert unreal_project_scenes == {"starter_content_0000", scene_id}
+        assert unreal_project_scenes == {"debug_0000", scene_id}
 
         # see https://docs.unrealengine.com/4.26/en-US/SharingAndReleasing/Deployment/Cooking for more information on these parameters
```
RachithP, line 131: Let's update this to https://docs.unrealengine.com/5.2/en-US/cooking-content-in-unreal-engine/ ?

cpp/unreal_projects/SpearSim/Source/SpearSimEditor/SpearSimEditorUtils.h
```diff
@@ -0,0 +1,34 @@
+//
```
RachithP, line 1: Why do we need this class? Are we using this in `debug_0000` scene?

mikeroberts3000: It's a proof-of-concept that shows how to add buttons and other UI elements to the Unreal editor in an extremely lightweight way. You can drag a SpearSimEditorUtils object from the content browser and drop it into the scene. Then a SpearSimEditorUtils object shows up in the World Outliner panel. When you click on the SpearSimEditorUtils object in the World Outliner, your custom UI shows up in the Details panel. One of these objects is currently placed in the debug_0000 scene.

RachithP (approved):
Looks great! 🥳

cpp/unreal_projects/SpearSim/Source/SpearSimEditor/SpearSimEditorUtils.h
mikeroberts3000: I'm envisioning a situation where some of our custom UI elements might contain per-user settings. We probably want each setting to have a default value, but we also want each user to be able to configure some of their settings individually and have the editor remember those per-user settings. This would be useful, e.g., if one of the UI elements was an output folder where we want the UI to save the NavMesh configurations that we're authoring. This should be a per-user setting, because each user may want to output to a different folder. (This is an imperfect example, because we will probably want to standardize and always save to a particular folder relative to the current uproject file, but it is still a generally useful idea to design a UI such that it remembers your individual settings.)

The DefaultSpear.ini file and corresponding C++ variable declarations provide exactly this behavior. A default value can be provided in DefaultSpear.ini, but the editor will remember your per-user value in a separate ini file.

tools/build_paks.py
mikeroberts3000: Sure

cpp/unreal_projects/SpearSim/Source/SpearSimEditor/SpearSimEditorUtils.h
RachithP: Makes sense. Where are `LoadConfig()` and `SaveConfig()` defined/implemented?

mikeroberts3000: I believe these functions are inherited from `UObject`. They load and save from whatever ini is implied by the class decorator (in this case `Spear.ini`). The load function will try to load from an per-user ini first, and fall back to the default ini (in this case `DefaultSpear.ini`) if a per-user ini is not found. The save function will save a per-user ini file, so it won't overwrite the default ini, unless it is configured to do otherwise (either with additional decorators on the UObject class or with additional function arguments).

PR #204, opened 2023-05-31
urdf for ue5

rwang15:


cpp/unreal_plugins/SimulationController/Source/SimulationController/UrdfBotAgent.h
```diff
@@ -20,7 +17,7 @@ class UWorld;
 class AUrdfBotPawn;
 class CameraSensor;
 class ImuSensor;
-class SonarSensor;
+//class SonarSensor;
```
mikeroberts3000, line 20: We decided recently to remove SonarSensor entirely until we have a chance to clean it up and generalize it into a general-purpose raycasting sensor.

cpp/unreal_plugins/UrdfBot/Source/UrdfBot/UrdfBot.Build.cs
```diff
@@ -10,7 +10,7 @@ public UrdfBot(ReadOnlyTargetRules target) : base(target)
     {
         SP_LOG_CURRENT_FUNCTION();
 
-        PublicDependencyModuleNames.AddRange(new string[] {"CoreUtils"});
+        PublicDependencyModuleNames.AddRange(new string[] {"CoreUtils", "XmlParser"});
```
mikeroberts3000, line 13: Does this need to be here or can it be moved to `CommonModuleRules` without breaking other plugins?

PR #207, opened 2023-06-21
Update generate_image_dataset example

RachithP:


examples/generate_image_dataset/generate_images.py
```diff
@@ -100,8 +100,8 @@ def single_step(self, action=None, get_observation=False):
 
         obs, _, _, _ = env.step(
             action={
-                "set_location": np.array([pose["location_x"], pose["location_y"], pose["location_z"]], np.float64),
-                "set_rotation": np.array([pose["rotation_pitch"], pose["rotation_yaw"], pose["rotation_roll"]], np.float64)})
+                "set_location": np.array([pose["location_x_cms"], pose["location_y_cms"], pose["location_z_cms"]], np.float64),
```
mikeroberts3000, line 103: I don't like this change. I think I have advocated for including units in interface names previously, so I realize that I'm going back and forth here. And I agree with the general principle of being specific with the names of things and making things clear and unambiguous for the reader.

However, I think an even cleaner convention is to simply pass values directly through to Unreal unless noted otherwise, and label our interfaces without units. In other words, I think it's better to just say `location_x`. If the reader is wondering what the units are, they can just remember our simple and consistent rule: _SPEAR always passes values directly through to Unreal unless noted otherwise, so units are whatever Unreal expects_.

I also think abbreviating centimeters as `cms` is confusing because it could be interpreted as `centimeters` or `centimeter*seconds`. The correct interpretation is pretty obvious in this case, but for other actions, scaling constants, etc throughout our code, the correct interpretation wouldn't always be so clear.

Also, if we labeled this column as `location_x_cms`, then I'd want to label tons of other things in our codebase with the correct units (e.g., dozens of config parameters, etc). That's too much effort and I'd rather be consistent. (There are a couple of parameters in `SPEAR` that are labeled with units, but those values aren't passed through to Unreal.)

Finally, as a matter of correctness, I don't think Unreal units are always centimeters. I think they're just centimeters by default. So the name `location_x_cms` might be misleading unless we're explicitly checking the `WorldToMetersScale` variable in Unreal.

examples/generate_image_dataset/generate_poses.py
```diff
@@ -42,33 +42,28 @@
...
+        # generate random pitch, yaw, roll values
+        pitch_values = np.random.uniform(low=0.0, high=0.0, size=args.num_poses_per_scene)
+        yaw_values   = np.random.uniform(low=-180.0, high=180.0, size=args.num_poses_per_scene)
+        roll_values  = np.random.uniform(low=0.0, high=0.0, size=args.num_poses_per_scene)
+
+        # store poses in a csv file
+        df = pd.DataFrame({"scene_id"            : scene_id,
+                           "location_x_cms"      : points[:,0],
+                           "location_y_cms"      : points[:,1],
+                           "location_z_cms"      : points[:,2] + config.SIMULATION_CONTROLLER.NAVMESH.AGENT_HEIGHT,
```
mikeroberts3000, line 57: I don't think we should use `config.SIMULATION_CONTROLLER.NAVMESH.AGENT_HEIGHT` for this purpose because it's kind of confusing and I don't want to convey the wrong idea to the reader. You can just hardcode a specific value here as a constant in this file, similar to `NUM_STEPS` in `getting_started.py`.

examples/generate_image_dataset/generate_poses.py
```diff
@@ -42,33 +42,28 @@
...
+                           "rotation_pitch_degs" : pitch_values,
+                           "rotation_yaw_degs"   : yaw_values,
+                           "rotation_roll_degs"  : roll_values})
+        df.to_csv(args.poses_file, mode="w" if i==0 else "a", index=False, header=i==0)
```
mikeroberts3000, line 61: I was never a fan of this approach for generating csv files. Just append to `df` and write it once. The current approach forces you to use `enumerate` and specify a bunch of extra named arguments to `to_csv`, which is slightly uglier than necessary.

examples/generate_image_dataset/user_config.yaml.example
```diff
@@ -19,6 +19,19 @@ SIMULATION_CONTROLLER:
   TASK: "NullTask"
   CUSTOM_UNREAL_CONSOLE_COMMANDS: []
 
+  NAVMESH:
```
mikeroberts3000, line 22: I don't think this navmesh stuff needs to be here because these are somewhat low-level default parameters that most users won't need to care about. My approach to including default parameters in `user_config.yaml.example` files so far is to include them only when I especially want to highlight the parameter. For example, the default for `FOV` is already `90.0` but I like that this config value is in `user_config.yaml.example` because it's helpful for the reader.

examples/generate_image_dataset/user_config.yaml.example
```diff
@@ -19,6 +19,19 @@ SIMULATION_CONTROLLER:
...
+    AGENT_RADIUS: 50.0
+    AGENT_HEIGHT: 200.0
+    AGENT_MAX_SLOPE: 1.0
+    AGENT_MAX_STEP_HEIGHT: 1.0
+    MIN_REGION_AREA: 0.0
+    MERGE_REGION_SIZE: 400.0
+    MAX_SIMPLIFICATION_ERROR: 1.3
+
   CAMERA_AGENT:
     CAMERA_ACTOR_NAME: "CameraActor"
```
mikeroberts3000, line 36: `"camera_actor"` to match our convention in `getting_started`.

mikeroberts3000 (approved):
Some minor stuff to fix but otherwise this looks great. We'll want to update the default scene to `apartment_0000` when its ready, but otherwise we've got another `example` in the bag.

examples/generate_image_dataset/generate_images.py
RachithP: That makes sense. I've updated the code accordingly.

examples/generate_image_dataset/generate_poses.py
RachithP: Updated

examples/generate_image_dataset/generate_poses.py
RachithP: Updated

examples/generate_image_dataset/user_config.yaml.example
RachithP: Removed all Navmesh default params (as set in default_config.simulation_controller.yaml)

examples/generate_image_dataset/user_config.yaml.example
RachithP: Updated

PR #208, opened 2023-07-07
Imitation learning example with UE 5.2

RachithP:


cpp/unreal_plugins/SimulationController/Source/SimulationController/ImitationLearningTask.cpp
```diff
@@ -9,8 +9,6 @@
 #include <string>
 #include <vector>
 
-#include <AI/Navigation/NavigationTypes.h>
-#include <AI/NavigationSystemBase.h>
 #include <Components/SceneComponent.h>
 #include <Delegates/IDelegateInstance.h>
 #include <DrawDebugHelpers.h>
```
mikeroberts3000, line 14: It is good that you have removed some headers but I think there are more that can be removed, e.g., `DebugDrawHelpers.h`.

cpp/unreal_plugins/SimulationController/Source/SimulationController/ImitationLearningTask.cpp
```diff
@@ -52,15 +46,17 @@ ImitationLearningTask::ImitationLearningTask(UWorld* world)
     actor_hit_event_component_->RegisterComponent();
     actor_hit_event_handle_ = actor_hit_event_component_->delegate_.AddRaw(this, &ImitationLearningTask::actorHitEventHandler);
 
-    // If the start/goal positions are not randomly generated, get them from a file
-    if (!Config::get<bool>("SIMULATION_CONTROLLER.IMITATION_LEARNING_TASK.GET_POSITIONS_FROM_TRAJECTORY_SAMPLING")) {
+    // Get start and end goal locations from a file
```
mikeroberts3000, line 49: - Don't make this file loading behavior optional. Make it mandatory, i.e., not wrapped in an if-statement.
- You can even inline the logic to load episodes directly into the constructor if you prefer. There is so little happening in this class now, so it won't be too confusing.
- The usage of "initial" vs "init" vs "start" vs "end" vs "goal" in this file is inconsistent and ugly. Pick a convention and stick to it throughout this entire class and the CSV file and the example code. I like "initial" and "goal".

cpp/unreal_plugins/SimulationController/Source/SimulationController/ImitationLearningTask.cpp
```diff
@@ -217,11 +189,11 @@ void ImitationLearningTask::getPositionsFromFile()
     position_index_ = -1;
 
     // Create an input filestream 
-    std::ifstream fs(Config::get<std::string>("SIMULATION_CONTROLLER.IMITATION_LEARNING_TASK.POSITIONS_FILE")); 
+    std::ifstream fs(Config::get<std::string>("SIMULATION_CONTROLLER.IMITATION_LEARNING_TASK.TRAJECTORY_LOCATIONS_FILE"));
```
mikeroberts3000, line 192: `EPISODES_FILE`

cpp/unreal_plugins/SimulationController/Source/SimulationController/ImitationLearningTask.cpp
```diff
@@ -244,102 +216,6 @@ void ImitationLearningTask::getPositionsFromFile()
     position_index_ = 0;
```
mikeroberts3000, line 216: `episode_index_`

cpp/unreal_plugins/SimulationController/Source/SimulationController/ImitationLearningTask.h
```diff
@@ -44,26 +42,18 @@ class ImitationLearningTask : public Task {
     void actorHitEventHandler(AActor* self_actor, AActor* other_actor, FVector normal_impulse, const FHitResult& hit_result);
 
     // Get a list of the different initial and goal positions the agent is expected to navigate to.
-    // Positions are in the format "init.X, init.Y, init.Z, goal.X, goal.Y, goal.Z".
+    // Positions are in the format "start_location.X, start_location.Y, start_location.Z, end_location.X, end_location.Y, end_location.Z".
```
mikeroberts3000, line 45: Need to include `scene_id` in this comment.

cpp/unreal_plugins/SimulationController/Source/SimulationController/NavMesh.h
```diff
@@ -19,7 +20,10 @@ class NavMesh
     void cleanUpObjectReferences();
 
     std::vector<uint8_t> getRandomPoints(int num_points);
-
+    std::vector<uint8_t> getReachablePoints(const std::vector<std::vector<float>>& start_points);
```
mikeroberts3000, line 23: I think for consistency it would be better if all of these functions in our `NavMesh` class used `std::vector<float>` for input and output. When I was initially designing the interface to `getRandomPoints()`, I was trying to be consistent with our existing gym interface, which uses `std::vector<uint8_t>`. But my initial thinking was misguided. My current thinking is as follows.

- First, our gym interface has the concept of spaces for describing the shape of input and output data. So it is always possible for surrounding Python code to correctly decode an output `std::vector<uint8_t>` without needing to know any internal C++ implementation details, based on the information contained in the corresponding space. But it feels like too much work to implement a different `getWhateverSpace()` function for every scene querying function we want (e.g., `getRandomPointsSpace()`, `getReachablePointsSpace()`, etc). However, without these `getWhateverSpace()` functions, surrounding Python code needs to know too much about internal C++ implementation details to decode the data correctly. We can see evidence of this unpleasantly tight coupling in my `Env.get_random_points()` function, which needs to know somehow that the underlying C++ code returns `np.float64` data, even though this contract is not specified anywhere in the interface, and you literally need to comb through the C++ implementation to figure that out.
- Second, we need the flexibility of `std::vector<uint8_t>` in our gym interface because our agents are totally general, supporting totally heterogeneous input and output data. But we don't need this flexibility in our scene querying interfaces because the shape of the input and output data is typically known ahead of time.
- Third, I believe there is a performance penalty for sending anything other than `std::vector<uint8_t>` using `rpclib`, but I don't think that is a big deal. Our scene querying interfaces will probably not be used in tight loops, so I think it is ok to have good enough performance in exchange for cleaner interfaces. And we can always implement a fast `std::vector<uint8_t> getRandomPointsAsUint8()` function if we really need maximum performance for some reason.

mikeroberts3000: - `getRandomReachablePoints`
- `getPaths()`

cpp/unreal_plugins/SimulationController/Source/SimulationController/NavMesh.h
```diff
@@ -19,7 +20,10 @@ class NavMesh
...
+    std::vector<std::vector<uint8_t>> getTrajectories(const std::vector<std::vector<float>>& start_points, const std::vector<std::vector<float>>& end_points);
 private:
```
mikeroberts3000, line 25:
```

private:
    UWorld* world_ = nullptr;
    UNavigationSystemV1* navigation_system_v1_ = nullptr;
    ARecastNavMesh* recast_nav_mesh_ = nullptr;
```

examples/getting_started/OpenBotEnv.py
```diff
@@ -0,0 +1,99 @@
+#
+# Copyright(c) 2022 Intel. Licensed under the MIT License <http://opensource.org/licenses/MIT>.
+#
+
+import gym
```
mikeroberts3000, line 5: Let's move this file into an `examples/common` folder and call it `openbot_env.py` for consistency with the existing `env.py` file. The rule "examples are totally self-contained except for the `common` folder" is a less confusing convention than "examples can depend on each other whenever we feel like it". Also, it is arbitrary that `OpenBotEnv` would live with `getting_started` instead of `imitation_learning_openbot`, so it is better to put it in `common` and then use a consistent hack in both examples to be able to import it.

cpp/unreal_plugins/SimulationController/Source/SimulationController/NavMesh.cpp
```diff
@@ -83,3 +90,72 @@ std::vector<uint8_t> NavMesh::getRandomPoints(int num_points)
     }
     return Std::reinterpretAs<uint8_t>(points);
 }
+
+std::vector<uint8_t> NavMesh::getReachablePoints(const std::vector<std::vector<float>>& start_points)
+{
+    std::vector<double> end_points;
```
mikeroberts3000, line 96: `std::vector<double> reachable_points`

cpp/unreal_plugins/SimulationController/Source/SimulationController/NavMesh.cpp
```diff
@@ -83,3 +90,72 @@ std::vector<uint8_t> NavMesh::getRandomPoints(int num_points)
     }
     return Std::reinterpretAs<uint8_t>(points);
 }
+
+std::vector<uint8_t> NavMesh::getReachablePoints(const std::vector<std::vector<float>>& start_points)
```
mikeroberts3000, line 94: `reference_points`

cpp/unreal_plugins/SimulationController/Source/SimulationController/NavMesh.cpp
```diff
@@ -83,3 +90,72 @@ std::vector<uint8_t> NavMesh::getRandomPoints(int num_points)
...
+{
+    std::vector<double> end_points;
+    for (auto& start_point : start_points) {
```
mikeroberts3000, line 97: Assert that `reference_points` is a multiple of 3, use an old-fashioned for-loop.

cpp/unreal_plugins/SimulationController/Source/SimulationController/NavMesh.cpp
```diff
@@ -83,3 +90,72 @@ std::vector<uint8_t> NavMesh::getRandomPoints(int num_points)
...
+
+        SP_ASSERT(start_point.size()==3);
+
+        FVector start_point_fvector{start_point.at(0), start_point.at(1), start_point.at(2)};
+        FNavLocation end_location;
+
+        bool found = recast_nav_mesh_->GetRandomReachablePointInRadius(
```
mikeroberts3000, line 104: The radius should be passed as input to this function, remove the config parameter from `default_config`.

cpp/unreal_plugins/SimulationController/Source/SimulationController/NavMesh.cpp
```diff
@@ -83,3 +90,72 @@ std::vector<uint8_t> NavMesh::getRandomPoints(int num_points)
...
 }
+
+std::vector<uint8_t> NavMesh::getReachablePoints(const std::vector<std::vector<float>>& start_points)
+{
+    std::vector<double> end_points;
+    for (auto& start_point : start_points) {
+
+        SP_ASSERT(start_point.size()==3);
+
+        FVector start_point_fvector{start_point.at(0), start_point.at(1), start_point.at(2)};
```
mikeroberts3000, line 101:
```
FVector reference_point = {reference_points.at(i), reference_points.at(i+1), reference_points.at(i+2)};
```

cpp/unreal_plugins/SimulationController/Source/SimulationController/NavMesh.cpp
```diff
@@ -83,3 +90,72 @@ std::vector<uint8_t> NavMesh::getRandomPoints(int num_points)
...
+    return Std::reinterpretAs<uint8_t>(end_points);
+}
+
+std::vector<std::vector<uint8_t>> NavMesh::getTrajectories(const std::vector<std::vector<float>>& start_points, const std::vector<std::vector<float>>& end_points)
+{
+    SP_ASSERT(start_points.size() == end_points.size());
+
+    std::vector<std::vector<uint8_t>> trajectories;
+
+    for (int i = 0; i < start_points.size(); ++i) {
```
mikeroberts3000, line 121: `i++`

cpp/unreal_plugins/SimulationController/Source/SimulationController/NavMesh.cpp
```diff
@@ -83,3 +90,72 @@ std::vector<uint8_t> NavMesh::getRandomPoints(int num_points)
...
+        end_points.push_back(end_location.Location.Z);
+    }
+    return Std::reinterpretAs<uint8_t>(end_points);
+}
+
+std::vector<std::vector<uint8_t>> NavMesh::getTrajectories(const std::vector<std::vector<float>>& start_points, const std::vector<std::vector<float>>& end_points)
+{
+    SP_ASSERT(start_points.size() == end_points.size());
+
+    std::vector<std::vector<uint8_t>> trajectories;
```
mikeroberts3000, line 119: `paths`

cpp/unreal_plugins/SimulationController/Source/SimulationController/NavMesh.cpp
```diff
@@ -83,3 +90,72 @@ std::vector<uint8_t> NavMesh::getRandomPoints(int num_points)
...
+        std::vector<double> trajectory;
+        for (auto& path_point : path_points) {
+            trajectory.push_back(path_point.Location.X);
+            trajectory.push_back(path_point.Location.Y);
+            trajectory.push_back(path_point.Location.Z);
+        }
+
+        trajectories.push_back(Std::reinterpretAs<uint8_t>(trajectory));
+
+        // Debug drawings
```
mikeroberts3000, line 151: Remove this debug drawing functionality. It's too application-specific for this general-purpose scene querying functionality. We're not really using it for anything currently. And I'd rather move this debug draw functionality into a general-purpose `DebugDraw` interface (similar to this `NavMesh` interface), where users could call from their Python code if they wanted to.

examples/getting_started/__init__.py
```diff
@@ -0,0 +1,3 @@
+#
```
mikeroberts3000, line 1: Do we need this if we're moving `openbot_env.py` into a `common` folder?

examples/imitation_learning_openbot/generate_dataset.py
```diff
@@ -11,12 +11,18 @@
 import os
 import pandas as pd
 import shutil
-import spear
```
mikeroberts3000, line 14: I like the `spear` import being up here. It is a "lower-level" import than these other example-specific imports.

examples/imitation_learning_openbot/generate_dataset.py
```diff
@@ -35,72 +40,52 @@
...
     config.defrost()
-    config.SIMULATION_CONTROLLER.IMITATION_LEARNING_TASK.GET_POSITIONS_FROM_TRAJECTORY_SAMPLING = False
-    config.SIMULATION_CONTROLLER.IMITATION_LEARNING_TASK.POSITIONS_FILE = os.path.abspath(args.episodes_file)
+    config.SIMULATION_CONTROLLER.IMITATION_LEARNING_TASK.LOAD_TRAJECTORY_FROM_FILE = True
+    config.SIMULATION_CONTROLLER.IMITATION_LEARNING_TASK.TRAJECTORY_LOCATIONS_FILE = os.path.abspath(args.episodes_file)
     config.freeze()
  
     # handle debug configuration (markers are only produed in Developent configuration; NOT in Shipping configuration)
+    config.defrost()
     if args.debug:
```
mikeroberts3000, line 51: - I'd rather have the default configuration live in user_config, and then only make changes if `args.debug` is true.
- I don't think RENDER_OFFSCREEN should be true in debug mode.

examples/imitation_learning_openbot/generate_dataset.py
```diff
@@ -122,65 +107,43 @@
...
 
             # create Env object
-            env = spear.Env(config=config)
+            env = OpenBotEnv(config=config)
                 
         # reset the simulation
         env_reset_info = {}
-        _ = env.reset(reset_info=env_reset_info)
+        obs = env.reset(reset_info=env_reset_info)
         assert "success" in env_reset_info
```
mikeroberts3000, line 119:
```
obs = env.reset(reset_info=env_reset_info)
assert env_reset_info["success"]
```

examples/imitation_learning_openbot/policies.py
```diff
@@ -5,17 +5,17 @@
...
 from utils import get_compass_observation, get_relative_target_pose
 
-
 class OpenBotPIDPolicy():
 
     def __init__(self, config): 
 
         self._config = config
-        self._trajectory = np.empty((0, 3), dtype=float)
+        self._trajectory = np.empty((0, 3), dtype=np.float64)
```
mikeroberts3000, line 16: `self._path`

examples/imitation_learning_openbot/policies.py
```diff
@@ -25,33 +25,33 @@ def __init__(self, config):
         assert self._config.IMITATION_LEARNING_OPENBOT.PID.PROPORTIONAL_GAIN_HEADING >= 0.0
         assert self._config.IMITATION_LEARNING_OPENBOT.PID.DERIVATIVE_GAIN_HEADING >= 0.0
         assert self._config.IMITATION_LEARNING_OPENBOT.PID.FORWARD_MIN_ANGLE >= 0.0
-        assert self._config.SIMULATION_CONTROLLER.SIMULATION_STEP_TIME_SECONDS > 0.0 
+        assert self._config.SIMULATION_CONTROLLER.SIMULATION_STEP_TIME > 0.0 
     
-    def reset(self, obs, env_step_info):
+    def reset(self, obs, trajectory):
```
mikeroberts3000, line 30: `path`

examples/imitation_learning_openbot/policies.py
```diff
@@ -25,33 +25,33 @@ def __init__(self, config):
...
         
-        self._trajectory = env_step_info["agent_step_info"]["trajectory_data"]
+        self._trajectory = trajectory[0]
```
mikeroberts3000, line 32: `self._path = path`

examples/imitation_learning_openbot/policies.py
```diff
@@ -25,33 +25,33 @@ def __init__(self, config):
...
+        xy_position_error, yaw_error = get_relative_target_pose(position_xy_desired, obs["location"][0:2], np.deg2rad(obs["rotation"][1]))
 
         # compute Euclidean distance to target in [m]:
         xy_position_error_norm = np.linalg.norm(xy_position_error) * 0.01
 
         # velocity computation
-        lin_vel_norm = np.linalg.norm((obs["state_data"][0:2] - self._position_xy_old) / self._config.SIMULATION_CONTROLLER.SIMULATION_STEP_TIME_SECONDS) * 0.01 
-        ang_vel_norm = (obs["state_data"][4] - self._yaw_old)/self._config.SIMULATION_CONTROLLER.SIMULATION_STEP_TIME_SECONDS
+        lin_vel_norm = np.linalg.norm((obs["location"][0:2] - self._position_xy_old) / self._config.SIMULATION_CONTROLLER.SIMULATION_STEP_TIME) * 0.01 
+        ang_vel_norm = (np.deg2rad(obs["rotation"][1]) - self._yaw_old)/self._config.SIMULATION_CONTROLLER.SIMULATION_STEP_TIME
```
mikeroberts3000, line 51: For consistency:

```
... = (...) / ...
```

examples/imitation_learning_openbot/policies.py
```diff
@@ -5,17 +5,17 @@
...
 
-
 class OpenBotPIDPolicy():
 
     def __init__(self, config): 
 
         self._config = config
-        self._trajectory = np.empty((0, 3), dtype=float)
+        self._trajectory = np.empty((0, 3), dtype=np.float64)
         self._index_waypoint = 0
```
mikeroberts3000, line 17: `self._waypoint_index`

examples/imitation_learning_openbot/policies.py
```diff
@@ -5,17 +5,17 @@
...
-        self._position_xy_old = np.zeros(2, dtype=np.float32)
+        self._position_xy_old = np.zeros(2, dtype=np.float64)
```
mikeroberts3000, line 18: Use "location" instead of "position" for consistency.

examples/imitation_learning_openbot/policies.py
```diff
@@ -25,33 +25,33 @@ def __init__(self, config):
...
 
     def step(self, obs):
-
+        
         # xy position of the next waypoint in world frame:
         position_xy_desired = self._trajectory[self._index_waypoint][0:2] # [x_des, y_des]
 
         # compute the relative agent-target pose from the raw observation dictionary
-        xy_position_error, yaw_error = get_relative_target_pose(position_xy_desired, obs["state_data"][0:2], obs["state_data"][4])
+        xy_position_error, yaw_error = get_relative_target_pose(position_xy_desired, obs["location"][0:2], np.deg2rad(obs["rotation"][1]))
```
mikeroberts3000, line 44:
```
cm_to_m = 0.01

location_xy_current = obs["location"][0:2] * cm_to_m
location_xy_desired = self._path[self._waypoint_index][0:2]
location_xy_error   = np.linalg.norm(location_xy_desired - location_xy_current)

rotation_yaw_current = np.deg2rad(obs["rotation"][1])
heading_xy_current   = np.array([np.cos(rotation_yaw_current), np.sin(rotation_yaw_current)])
heading_xy_desired   = (location_xy_desired - location_xy_current) / np.linalg.norm(location_xy_desired - location_xy_current)
rotation_yaw_error   = np.arctan2(heading_xy_desired[1], heading_xy_desired[0]) - np.arctan2(heading_xy_current[1], heading_xy_current[0])

if rotation_yaw_error < -np.pi:
    rotation_yaw_error += 2*np.pi
if rotation_yaw_error > np.pi:
    rotation_yaw_error -= 2*np.pi

location_xy_prev  = self._location_xy_prev
rotation_yaw_prev = self._rotation_yaw_prev

location_xy_speed  = np.linalg.norm(location_xy_current - location_xy_prev) / self._config.SIMULATION_CONTROLLER.SIMULATION_STEP_TIME_SECONDS)
rotation_yaw_speed = np.linalg.norm(rotation_yaw_current - rotation_yaw_prev) / self._config.SIMULATION_CONTROLLER.SIMULATION_STEP_TIME_SECONDS)
```

mikeroberts3000:
```
# multiply by -1.0 because otherwise a positive rotation error would cause us to turn right, but it should cause us to turn left
control_right = \
    -1.0 *
    ((rotation_yaw_error * self._config.IMITATION_LEARNING_OPENBOT.ROTATION_YAW_PROPORTIONAL_GAIN) + \
    (rotation_yaw_speed * self._config.IMITATION_LEARNING_OPENBOT.ROTATION_YAW_DERIVATIVE_GAIN))

# multiply by an exponential decay term to prioritize rotation errors, use LOCATION_XY_HALF_LIFE to control how agressively to prioritize rotation errors
control_forward = \
    np.pow(2.0, -abs(rotation_yaw_error) / self._config.IMITATION_LEARNING_OPENBOT.LOCATION_XY_HALF_LIFE) * \ 
    ((location_xy_error * self._config.IMITATION_LEARNING_OPENBOT.LOCATION_XY_PROPORTIONAL_GAIN) + \
    (location_xy_speed * self._config.IMITATION_LEARNING_OPENBOT.LOCATION_XY_DERIVATIVE_GAIN))

action = np.array([control_forward + control_right, control_forward - control_left])
action = np.clip(action, -self._config.IMITATION_LEARNING_OPENBOT.MAX_ACTION, self._config.IMITATION_LEARNING_OPENBOT.MAX_ACTION)

```

examples/imitation_learning_openbot/policies.py
```diff
@@ -66,17 +66,17 @@ def step(self, obs):
         left_wheel_command = np.clip(left_wheel_command, -self._config.IMITATION_LEARNING_OPENBOT.CONTROL_SATURATION, self._config.IMITATION_LEARNING_OPENBOT.CONTROL_SATURATION)
         right_wheel_command = forward_ctrl - right_ctrl
         right_wheel_command = np.clip(right_wheel_command, -self._config.IMITATION_LEARNING_OPENBOT.CONTROL_SATURATION, self._config.IMITATION_LEARNING_OPENBOT.CONTROL_SATURATION)
-        action = np.array([left_wheel_command,right_wheel_command], dtype=np.float32)
+        action = np.array([left_wheel_command,right_wheel_command], dtype=np.float64)
 
         # update member variables
-        self._position_xy_old = obs["state_data"][0:2]
-        self._yaw_old = obs["state_data"][4]
+        self._position_xy_old = obs["location"][0:2]
```
mikeroberts3000, line 72:
```
self._location_xy_prev = location_xy_current
self._rotation_yaw_prev = rotation_yaw_prev

 waypoint_reached = location_xy_error <= self._config.IMITATION_LEARNING_OPENBOT.ACCEPTANCE_RADIUS)

if waypoint_reached:
    num_waypoints = self._path.shape[0]
    if self._index_waypoint < num_waypoints - 1:
        # ...
```
```

File: examples/imitation_learning_openbot/policies.py [hunk 29]
```diff
@@ -5,17 +5,17 @@
 import numpy as np
 import tflite_runtime.interpreter as tflite
 
+import spear
 from utils import get_compass_observation, get_relative_target_pose
 
-
 class OpenBotPIDPolicy():
```
mikeroberts3000, line 11: `OpenBotWaypointFollowingPolicy` because it isn't really implementing a PID controller.

examples/imitation_learning_openbot/generate_episodes.py
```diff
@@ -16,22 +16,23 @@
 
     parser = argparse.ArgumentParser()
     parser.add_argument("--num_episodes_per_scene", type=int, default=10)
+    parser.add_argument("--num_candidate_points_per_episode", type=int, default=10)
```
mikeroberts3000, line 19: `num_candidates_per_episode`

examples/imitation_learning_openbot/generate_episodes.py
```diff
@@ -16,22 +16,23 @@
...
     parser.add_argument("--episodes_file", default=os.path.realpath(os.path.join(os.path.dirname(__file__), "train_episodes.csv")))
     parser.add_argument("--scene_id")
     args = parser.parse_args()
 
-    # directory for trajectory image storage
-    split_name = os.path.splitext(os.path.basename(args.episodes_file))[0] # isolate filename and remove extension
+    # create directory for storing episodes information
```
mikeroberts3000, line 24: "episode information"

examples/imitation_learning_openbot/generate_episodes.py
```diff
@@ -42,71 +43,73 @@
...
 
-            # warehouse_0000 has scene-specific config values
-            scene_config_file = os.path.realpath(os.path.join(os.path.dirname(__file__), "scene_config.warehouse_0000.yaml"))
+        # reset the simulation
+        env_reset_info = {}
+        _ = env.reset(reset_info=env_reset_info)
+        assert "success" in env_reset_info
+            
+        # generate candidate points based out of args.num_episodes_per_scene
+        candidate_points = env.get_random_points(args.num_episodes_per_scene * args.num_candidate_points_per_episode)
```
mikeroberts3000, line 69: `candidate_initial_points`

examples/imitation_learning_openbot/generate_episodes.py
```diff
@@ -42,71 +43,73 @@
...
 
-        else:
-            assert False
+        # obtain a reachable goal point for every candidate point
+        reachable_points = env.get_reachable_points(candidate_points.tolist())
```
mikeroberts3000, line 72: - `candidate_goal_points`
-  do the `tolist()` conversion inside the function for consistency with all our other python functions functions that assume numpy arrays as an input/output format

examples/imitation_learning_openbot/generate_episodes.py
```diff
@@ -42,71 +43,73 @@
...
 
-        config.merge_from_file(scene_config_file)
-        config.freeze()
+        # get trajectories for every candidate_points and corresponding reachable_points
```
mikeroberts3000, line 74: - `paths = get_paths(...)`

examples/imitation_learning_openbot/generate_episodes.py
```diff
@@ -42,71 +43,73 @@
...
+                  "start_location_x" : merged_array[:,0],
+                  "start_location_y" : merged_array[:,1],
+                  "start_location_z" : merged_array[:,2],
+                  "goal_location_x"  : merged_array[:,3],
+                  "goal_location_y"  : merged_array[:,4],
+                  "goal_location_z"  : merged_array[:,5]})
+        
+        df = pd.concat([df, df_])
 
+        plt.plot(merged_array[:,0], merged_array[:,1], "^", markersize=12.0, label="Start", color="tab:blue", alpha=0.3)
```
mikeroberts3000, line 109: move this plotting code into `utils.py` for consistency with the other plotting code

examples/imitation_learning_openbot/generate_dataset.py
```diff
@@ -35,72 +40,52 @@
...
-        rendering_mode_map_str = "_rtx"
-        config.defrost()
-        config.SIMULATION_CONTROLLER.CAMERA_SENSOR.FINAL_COLOR_INDIRECT_LIGHTING_INTENSITY = 0.0
-        config.freeze()
-    else:
-        assert False
-        
-    # clean the episode data folder 
+
+    # clean the episode data folder
```
mikeroberts3000, line 88:
```
spear.log(f"Generating data for episode {episode['index']} of {df.shape[0]}...")
```

examples/imitation_learning_openbot/generate_dataset.py
```diff
@@ -122,65 +107,43 @@
...
-            spear.log("Call to env.reset(...) was not successful. Simulation took too long to return to a ready state. Skipping...")
-            prev_scene_id = episode["scene_id"]
-            continue
-
-        # send zero action to the agent and collect initial trajectory observations:
-        obs, _, _, env_step_info = env.step(action={"apply_voltage": np.array([0.0, 0.0], dtype=np.float32)})
+        # get a trajectory for this episode based on start and end point
+        episode_start_location = [[episode["start_location_x"], episode["start_location_y"], episode["start_location_z"]]]
+        episode_goal_location  = [[episode["goal_location_x"], episode["goal_location_y"], episode["goal_location_z"]]]
+        trajectory = env.get_trajectories(episode_start_location, episode_goal_location)
```
mikeroberts3000, line 124:
```
path = env.get_paths(episode_start_location, episode_goal_location)[0]
```

examples/imitation_learning_openbot/generate_dataset.py
```diff
@@ -122,65 +107,43 @@
...
-            waypoint_data = np.empty([args.num_iterations_per_episode, 3], dtype=np.float32) # waypoints being tracked by the agent during an episode
-            compass_data  = np.empty([args.num_iterations_per_episode, 3], dtype=np.float32) # compass observations made by the agent during a episode
+            control_data  = np.empty([args.num_iterations_per_episode, 2], dtype=np.float64) # control_data observations made by the agent during a episode
+            state_data    = np.empty([args.num_iterations_per_episode, 6], dtype=np.float64) # state_data observations made by the agent during an episode
+            waypoint_data = np.empty([args.num_iterations_per_episode, 3], dtype=np.float64) # waypoints being tracked by the agent during an episode
+            compass_data  = np.empty([args.num_iterations_per_episode, 3], dtype=np.float64) # compass observations made by the agent during a episode
             time_data     = np.empty([args.num_iterations_per_episode], dtype=np.int64)      # time stamps of observations made by the agent during an episode
             frame_data    = np.empty([args.num_iterations_per_episode], dtype=np.int32)      # frame ids
```
mikeroberts3000, line 149: `num_iterations_executed`

examples/imitation_learning_openbot/generate_dataset.py
```diff
@@ -258,7 +221,7 @@
         # get the updated compass observation (with the last recorded position set as goal)
         goal_position_xy = state_data[num_iterations-1][0:2]
         for i in range(num_iterations):
-            position_xy_current = np.array([state_data[i][0], state_data[i][1]], dtype=np.float32)
+            position_xy_current = np.array([state_data[i][0], state_data[i][1]], dtype=np.float64)
```
mikeroberts3000, line 224:
```
location_xy_desired = state_data[num_iterations_executed-1][0:2]
for i in range(num_iterations_executed):
    cm_to_m = 0.01
    
    location_xy_current = state_data[i, 0:2]
    location_xy_error   = np.linalg.norm(location_xy_desired - location_xy_current)
    
    rotation_yaw_current = state_data[i, 1]
    heading_xy_current   = np.array([np.cos(rotation_yaw_current), np.sin(rotation_yaw_current)])
    heading_xy_desired   = (location_xy_desired - location_xy_current) / np.linalg.norm(location_xy_desired - location_xy_current)
    rotation_yaw_error   = np.arctan2(heading_xy_desired[1], heading_xy_desired[0]) - np.arctan2(heading_xy_current[1], heading_xy_current[0])
    
    if rotation_yaw_error < -np.pi:
        rotation_yaw_error += 2*np.pi
    if rotation_yaw_error > np.pi:
        rotation_yaw_error -= 2*np.pi

    compass_data[i] = np.array([location_xy_error*cm_to_m, np.sin(rotation_yaw_error), np.cos(rotation_yaw_error)])

df_goal = pd.DataFrame({"timestamp[ns]" : time_data[:num_iterations_executed],
                        "dist[m]"       : compass_data[:num_iterations_executed, 0],
                        "sinYaw"        : compass_data[:num_iterations_executed, 1],
                        "cosYaw"        : compass_data[:num_iterations_executed, 2]})
df_goal.to_csv(os.path.realpath(os.path.join(sensor_dir, "goalLog.txt")), mode="w", index=False, header=True)
```

examples/imitation_learning_openbot/generate_dataset.py
```diff
@@ -217,16 +180,16 @@
                 # the new goal position. Doing so requires a recomputation of the compass observation, since the latter is goal dependant. Therefore, rather
                 # than directly writing all the observations in a file iteration by iteration, we append these observations in a buffer, named "observation"
                 # to later process them once the episode is completed. 
-                control_data[i]     = obs["control_data"]                  # control_data: [ctrl_left, ctrl_right]
-                state_data[i]       = obs["state_data"]                    # state_data: [x, y, z, pitch, yaw, roll]
-                waypoint_data[i]    = policy_step_info["current_waypoint"] # current waypoint being tracked by the agent
-                time_data[i]        = time_stamp                           # current time stamp
-                frame_data[i]       = i                                    # current frame
+                control_data[i]     = action                                                                    # control_data: [ctrl_left, ctrl_right]
```
mikeroberts3000, line 183:
```
timestamp_data[i]   = timestamp
frame_index_data[i] = i
control_data[i]     = action
debug_data[i]       = np.hstack([obs["location"], np.deg2rad(obs["rotation"]), policy_step_info["current_waypoint"]])

File: examples/imitation_learning_openbot/generate_dataset.py [hunk 41]
```diff
@@ -122,65 +107,43 @@
...
+            plots_dir   = os.path.realpath(os.path.join(episode_dir, "plots"))
             os.makedirs(image_dir, exist_ok=True)
             os.makedirs(sensor_dir, exist_ok=True)
             os.makedirs(plots_dir, exist_ok=True)
 
-            control_data  = np.empty([args.num_iterations_per_episode, 2], dtype=np.float32) # control_data observations made by the agent during a episode
-            state_data    = np.empty([args.num_iterations_per_episode, 6], dtype=np.float32) # state_data observations made by the agent during an episode
-            waypoint_data = np.empty([args.num_iterations_per_episode, 3], dtype=np.float32) # waypoints being tracked by the agent during an episode
-            compass_data  = np.empty([args.num_iterations_per_episode, 3], dtype=np.float32) # compass observations made by the agent during a episode
+            control_data  = np.empty([args.num_iterations_per_episode, 2], dtype=np.float64) # control_data observations made by the agent during a episode
```
mikeroberts3000, line 143:
```
episode_timestamp_data = ...
episode_frame_id_data  = ...
episode_control_data   = ...
episode_command_data   = ...
episode_debug_data     = ...
```

examples/imitation_learning_openbot/generate_dataset.py
```diff
@@ -258,7 +221,7 @@
         # get the updated compass observation (with the last recorded position set as goal)
         goal_position_xy = state_data[num_iterations-1][0:2]
         for i in range(num_iterations):
-            position_xy_current = np.array([state_data[i][0], state_data[i][1]], dtype=np.float32)
+            position_xy_current = np.array([state_data[i][0], state_data[i][1]], dtype=np.float64)
             yaw_current = state_data[i][4]
             compass_data[i] = get_compass_observation(goal_position_xy, position_xy_current, yaw_current)
```
mikeroberts3000, line 227: - write in this order: `ctrlLog.txt`, `rgbFrames.txt`, `goalLog.txt`, `debugLog.txt`
- combine all of the debug data into a single DataFrame

examples/imitation_learning_openbot/policies.py
```diff
@@ -121,23 +121,19 @@ def step(self, obs, position_xy_desired):
         for input_detail in self._input_details:
```
mikeroberts3000, line 121: Why are we checking for `"cmd_input"` or `"goal_input"` in this if statement? Are both present, and we pass in the same data for each? That seems weird.

```
        for input_detail in self._input_details:
            if "img_input" in input_detail["name"]:

                # select the bottom right corner of obs, normalize, convert to np.float32
                img_input_height = input_detail["shape"][1] 
                img_input_width  = input_detail["shape"][2] 
                obs_height       = obs["camera.final_color"].shape[0]
                obs_width        = obs["camera.final_color"].shape[1]

                assert obs_height - img_input_height >= 0
                assert obs_width - img_input_width   >= 0

                img_input = obs["camera.final_color"][np.newaxis, image_height-target_height:, image_width-target_width:, :-1].astype(np.float32) / 255.0
                self._interpreter.set_tensor(input_detail["index"], img_input)

            elif "cmd_input" in input_detail["name"] or "goal_input" in input_detail["name"]:

                # compute a goal observation
                cm_to_m = 0.01
                
                location_xy_current = obs["location"][0:2] * cm_to_m
                location_xy_error   = np.linalg.norm(location_xy_desired - location_xy_current)
                
                rotation_yaw_current = np.deg2rad(obs["rotation"][1])
                heading_xy_current   = np.array([np.cos(rotation_yaw_current), np.sin(rotation_yaw_current)])
                heading_xy_desired   = (location_xy_desired - location_xy_current) / np.linalg.norm(location_xy_desired - location_xy_current)
                rotation_yaw_error   = np.arctan2(heading_xy_desired[1], heading_xy_desired[0]) - np.arctan2(heading_xy_current[1], heading_xy_current[0])
                
                if rotation_yaw_error < -np.pi:
                    rotation_yaw_error += 2*np.pi
                if rotation_yaw_error > np.pi:
                    rotation_yaw_error -= 2*np.pi

                goal = np.array([location_xy_error, np.sin(rotation_yaw_error), np.cos(rotation_yaw_error)])
                self._interpreter.set_tensor(input_detail["index"], goal[np.newaxis])

            else:
                assert False
```

examples/imitation_learning_openbot/run_trained_policy.py
```diff
@@ -185,7 +150,7 @@
             action, policy_step_info = policy.step(obs, goal[0:2])
```
mikeroberts3000, line 150:
```
action, policy_step_info = policy.step(obs)
```

examples/imitation_learning_openbot/run_trained_policy.py
```diff
@@ -114,69 +101,47 @@
...
-            spear.log("Call to env.reset(...) was not successful. Simulation took too long to return to a ready state. Skipping...")
-            prev_scene_id = episode["scene_id"]
-            continue
-
-        # send zero action to the agent and collect initial trajectory observations:
-        obs, _, _, env_step_info = env.step(action={"apply_voltage": np.array([0.0, 0.0], dtype=np.float32)})
+        # get a trajectory for this episode based on start and end point
+        episode_start_location = [[episode["start_location_x"], episode["start_location_y"], episode["start_location_z"]]]
+        episode_goal_location  = [[episode["goal_location_x"], episode["goal_location_y"], episode["goal_location_z"]]]
+        trajectory = env.get_trajectories(episode_start_location, episode_goal_location)
```
mikeroberts3000, line 119: Add a `policy.reset(goal)` function here to pass in the high-level conditioning data (the goal position), similar to how we do it in `OpenBotPathFollowingPolicy`

examples/imitation_learning_openbot/utils.py
```diff
@@ -102,14 +103,14 @@ def generate_video(image_dir, video_path, rate, compress=False):
     process.wait()
 
 
-def plot_tracking_performance_spatial(poses_current, poses_desired, plot_path):
+def plot_tracking_performance_spatial(poses_tracked, poses_desired, plot_path):
```
mikeroberts3000, line 106: `poses_actual`

examples/imitation_learning_openbot/utils.py
```diff
@@ -135,26 +136,26 @@ def plot_tracking_performance_spatial(poses_current, poses_desired, plot_path):
     plt.savefig(plot_path, bbox_extra_artists=[legend], bbox_inches="tight")
 
 
-def plot_tracking_performance_temporal(poses_current, poses_desired, plot_path):
+def plot_tracking_performance_temporal(poses_tracked, poses_desired, plot_path):
```
mikeroberts3000, line 139: `poses_actual`

python/spear/env.py
```diff
@@ -442,6 +460,22 @@ def _get_random_points(self, num_points):
...
+        reachable_points_serialized = self._rpc_client.call("get_reachable_points", start_points)
+        reachable_points = _deserialize_array(
+            reachable_points_serialized, space=Box(low=-np.inf, high=np.inf, shape=(-1,3), dtype=np.float64), byte_order=self._byte_order)
+        return reachable_points
+
+    def _get_trajectories(self, start_points, end_points):
+        assert all(isinstance(i, list) for i in start_points)
+        assert all(isinstance(i, list) for i in end_points)
+        trajectories_serialized_list = self._rpc_client.call("get_trajectories", start_points, end_points)
+        trajectories = np.vectorize(
```
mikeroberts3000, line 474: Why is this better than a vanilla list comprehension?

examples/imitation_learning_openbot/utils.py
```diff
@@ -7,6 +7,7 @@
 import matplotlib.pyplot as plt
 import numpy as np
 import os
+import spear
 
 
 def show_obs(obs, obs_components, render_passes):
```
mikeroberts3000, line 13: Function not up-to-date.

examples/getting_started/OpenBotEnv.py
```diff
@@ -0,0 +1,99 @@
...
+        # In this derived class, we are expecting different actions than the base spear.Env class. So
+        # we need to override self.action_space. We need to do this after calling super().__init__(...),
+        # because the base spear.Env class will set self.action_space to the lower-level actions it is
+        # expecting internally.
+        self.action_space = gym.spaces.Dict(spaces={"set_duty_cycles": gym.spaces.Box(low=-1, high=1, shape=(2,), dtype=np.float64)})
+    
+    def reset(self, reset_info=None):
+    
+        obs = super().reset(reset_info=reset_info)
+        assert "wheel_encoder" in obs.keys()
```
mikeroberts3000, line 32: It would be better if the C++ observation component was called `"wheel_rotation_speeds"` instead of `"wheel_encoder"`.

examples/imitation_learning_openbot/generate_episodes.py
```diff
@@ -16,22 +16,23 @@
...
     args = parser.parse_args()
 
-    # directory for trajectory image storage
-    split_name = os.path.splitext(os.path.basename(args.episodes_file))[0] # isolate filename and remove extension
+    # create directory for storing episodes information
+    file_name = os.path.splitext(os.path.basename(args.episodes_file))[0] # isolate filename and remove extension
     episodes_dir = os.path.realpath(os.path.join(os.path.dirname(__file__), "episodes"))
-    split_dir = os.path.realpath(os.path.join(episodes_dir, split_name))
-    os.makedirs(split_dir, exist_ok=True)
+    output_dir = os.path.realpath(os.path.join(episodes_dir, file_name))
```
mikeroberts3000, line 27: - for simplicity, just write the the plots directly to `episodes_dir`

examples/imitation_learning_openbot/generate_dataset.py
```diff
@@ -312,7 +275,7 @@
             generate_video(
```
mikeroberts3000, line 275: Just use `video_dir` directly:

```
video_dir = os.path.realpath(os.path.join(args.dataset_dir, "videos", args.split + "_data", episode["scene_id"])
```

examples/imitation_learning_openbot/run_trained_policy.py
```diff
@@ -79,20 +80,6 @@
...
-        rendering_mode_map_str = "_rtx"
-        config.defrost()
-        config.SIMULATION_CONTROLLER.CAMERA_SENSOR.FINAL_COLOR_INDIRECT_LIGHTING_INTENSITY = 0.0
-        config.freeze()
-    else:
-        assert False
-
     # clean the episode data folder 
     if not args.benchmark:
         shutil.rmtree(args.eval_dir, ignore_errors=True) # remove the previous dataset to prevent data corruption
```
mikeroberts3000, line 85: Make log statement consistent with `generate_dataset.py`

examples/imitation_learning_openbot/generate_dataset.py
```diff
@@ -35,72 +40,52 @@
...
-        config.freeze()
-    else:
-        assert False
-        
-    # clean the episode data folder 
+
+    # clean the episode data folder
     if not args.benchmark:
         shutil.rmtree(args.dataset_dir, ignore_errors=True) # remove the previous dataset to prevent data corruption
         split_dir = os.path.realpath(os.path.join(args.dataset_dir, args.split + "_data"))
```
mikeroberts3000, line 91: Don't compute `split_dir` here because it isn't used until later.

examples/imitation_learning_openbot/generate_dataset.py
```diff
@@ -122,65 +107,43 @@
...
-        # initialize the driving policy with the desired trajectory 
-        policy.reset(obs, env_step_info)
+        # initialize the driving policy with the desired trajectory
+        policy.reset(obs, trajectory)
 
         if args.benchmark:
             start_time_seconds = time.time()
         else:     
             # build the episode data folder and its subfolders following the guidelines of the OpenBot public repository 
             # https://github.com/isl-org/OpenBot/tree/master/policy#data-collection
```
mikeroberts3000, line 133: Compute `split_dir` here.

examples/imitation_learning_openbot/generate_dataset.py
```diff
@@ -122,65 +107,43 @@
...
-            scene_dir = os.path.realpath(os.path.join(split_dir, episode["scene_id"]))
+            scene_dir   = os.path.realpath(os.path.join(split_dir, episode["scene_id"]))
```
mikeroberts3000, line 134:
```
scene_dir = os.path.realpath(os.path.join(args.dataset_dir, args.split + "_data", episode["scene_id"]))
```

examples/imitation_learning_openbot/generate_dataset.py
```diff
@@ -122,65 +107,43 @@
...
             episode_dir = os.path.realpath(os.path.join(scene_dir, "%04d" % episode["index"]))
-            image_dir = os.path.realpath(os.path.join(episode_dir, "images"))
-            sensor_dir = os.path.realpath(os.path.join(episode_dir, "sensor_data"))
-            plots_dir = os.path.realpath(os.path.join(episode_dir, "plots"))
+            image_dir   = os.path.realpath(os.path.join(episode_dir, "images"))
```
mikeroberts3000, line 136: `images_dir`, `sensor_data_dir`

examples/imitation_learning_openbot/run_trained_policy.py
```diff
@@ -114,69 +101,47 @@
...
             start_time_seconds = time.time()
         else:
             # create dirs for storing data
-            scene_dir = os.path.realpath(os.path.join(args.eval_dir, episode["scene_id"]))
+            scene_dir   = os.path.realpath(os.path.join(args.eval_dir, episode["scene_id"]))
             episode_dir = os.path.realpath(os.path.join(scene_dir, "%04d" % episode["index"]))
-            image_dir = os.path.realpath(os.path.join(episode_dir, "images"))
-            result_dir = os.path.realpath(os.path.join(episode_dir, "results"))
-            plots_dir = os.path.realpath(os.path.join(episode_dir, "plots"))
+            image_dir   = os.path.realpath(os.path.join(episode_dir, "images"))
```
mikeroberts3000, line 126: `images_dir`

examples/imitation_learning_openbot/run_trained_policy.py
```diff
@@ -114,69 +101,47 @@
...
 
             # buffer containing the state_data observations made by the agent during an episode
-            state_data = np.empty([args.num_iterations_per_episode, 6], dtype=np.float32)
+            state_data = np.empty([args.num_iterations_per_episode, 6], dtype=np.float64)
 
             # save the optimal goal trajectory in a dedicated file
-            df_trajectory = pd.DataFrame({"x_d[cm]" : [env_step_info["agent_step_info"]["trajectory_data"][:,0]],
-                                          "y_d[cm]" : [env_step_info["agent_step_info"]["trajectory_data"][:,1]],
-                                          "z_d[cm]" : [env_step_info["agent_step_info"]["trajectory_data"][:,2]]})
+            df_trajectory = pd.DataFrame({"x_d[cm]" : trajectory[0][:,0],
```
mikeroberts3000, line 137: - mimic the approach in `generate_dataset.py`, where we write the `DataFrames` together, and we do once at the end of each episode

examples/imitation_learning_openbot/generate_episodes.py
```diff
@@ -115,11 +118,14 @@
         plt.xlabel("x[cm]")
         plt.ylabel("y[cm]")
         plt.grid()
-        plt.title(f"scene_id {scene_id}")
+        plt.title(f"scene_id: {scene_id}")
         
-        plt.savefig(os.path.realpath(os.path.join(split_dir, scene_id)), bbox_extra_artists=[legend], bbox_inches="tight")
+        plt.savefig(os.path.realpath(os.path.join(output_dir, scene_id)), bbox_extra_artists=[legend], bbox_inches="tight")
```
mikeroberts3000, line 123: - create a `plot_file` variable here for the filename for readability

examples/imitation_learning_openbot/run_trained_policy.py
```diff
@@ -114,69 +101,47 @@
...
+            image_dir   = os.path.realpath(os.path.join(episode_dir, "images"))
+            result_dir  = os.path.realpath(os.path.join(episode_dir, "results"))
+            plots_dir   = os.path.realpath(os.path.join(episode_dir, "plots"))
             os.makedirs(image_dir, exist_ok=True)
             os.makedirs(result_dir, exist_ok=True)
             os.makedirs(plots_dir, exist_ok=True)
 
             # buffer containing the state_data observations made by the agent during an episode
-            state_data = np.empty([args.num_iterations_per_episode, 6], dtype=np.float32)
+            state_data = np.empty([args.num_iterations_per_episode, 6], dtype=np.float64)
```
mikeroberts3000, line 134: No need for `state_data` here

examples/imitation_learning_openbot/run_trained_policy.py
```diff
@@ -200,15 +165,15 @@
                 plt.imsave(os.path.realpath(os.path.join(image_dir, "%04d.jpg"%i)), obs_final_color)
 
                 # populate buffer and result data file
-                state_data[i] = obs["state_data"]
+                state_data[i] = np.concatenate((obs["location"], np.deg2rad(obs["rotation"])), axis=None)
                 df_result = pd.DataFrame({"left_ctrl"    : action[0],
```
mikeroberts3000, line 169: - `df_path_actual`
- convert to meters and radians
- fill in the data frame directly from `obs`, no need to construct an intermediate `state_data` variable

examples/imitation_learning_openbot/run_trained_policy.py
```diff
@@ -244,7 +209,7 @@
         # create plots
```
mikeroberts3000, line 209:
```
        # Create plots. Note that creating these plots will resize our cv2 windows in an
        # unpleasant way, so we only generate these plots if we're not in debug mode.
        if not args.debug:
            # ...
```

examples/imitation_learning_openbot/utils.py
```diff
@@ -102,14 +103,14 @@ def generate_video(image_dir, video_path, rate, compress=False):
...
     fig, (ax) = plt.subplots(1, 1)
 
-    ax.plot(poses_current[0,0], poses_current[0,1], marker="^", markersize=12.0, label="Start", color="tab:blue")
+    ax.plot(poses_tracked[0,0], poses_tracked[0,1], marker="^", markersize=12.0, label="Start", color="tab:blue")
     ax.plot(poses_desired[-1,0], poses_desired[-1,1], marker="^", markersize=12.0, label="Goal", color="tab:orange")
     ax.plot(poses_desired[:,0], poses_desired[:,1], marker="o", markersize=8.0, label="Desired trajectory", color="tab:green")
-    ax.plot(poses_current[:,0], poses_current[:,1], marker="x", markersize=3.0, label="Actual trajectory", color="tab:purple")
+    ax.plot(poses_tracked[:,0], poses_tracked[:,1], marker="x", markersize=3.0, label="Actual trajectory", color="tab:purple")
 
     x_0,x_1 = ax.get_xlim()
```
mikeroberts3000, line 115: It would be cleaner if these plotting functions assumed meters rather than centimeters.

examples/imitation_learning_openbot/run_trained_policy.py
```diff
@@ -114,69 +101,47 @@
...
-                                          "z_d[cm]" : [env_step_info["agent_step_info"]["trajectory_data"][:,2]]})
+            df_trajectory = pd.DataFrame({"x_d[cm]" : trajectory[0][:,0],
+                                          "y_d[cm]" : trajectory[0][:,1],
+                                          "z_d[cm]" : trajectory[0][:,2]})
             df_trajectory.to_csv(os.path.realpath(os.path.join(result_dir,"trajectoryLog.txt")), mode="w", index=False, header=True)
 
         # execute the desired number of iterations in a given episode
         num_iterations = 0
-        goal           = np.array([episode["goal_pos_x_cms"], episode["goal_pos_y_cms"], episode["goal_pos_z_cms"]], dtype=np.float32) # goal position
+        goal = np.array([episode["goal_location_x"], episode["goal_location_y"], episode["goal_location_z"]], dtype=np.float64) # goal position
```
mikeroberts3000, line 144: - this variable is redundant wrt `episode_goal_location`

examples/imitation_learning_openbot/run_trained_policy.py
```diff
@@ -220,8 +185,8 @@
             if args.debug:
                 show_obs(
                     obs,
-                    config.SIMULATION_CONTROLLER.OPENBOT_AGENT.OBSERVATION_COMPONENTS,
-                    config.SIMULATION_CONTROLLER.OPENBOT_AGENT.CAMERA.RENDER_PASSES)
+                    config.SIMULATION_CONTROLLER.VEHICLE_AGENT.OBSERVATION_COMPONENTS,
+                    config.SIMULATION_CONTROLLER.VEHICLE_AGENT.CAMERA.RENDER_PASSES)
             
             # termination conditions
             if env_step_info["task_step_info"]["hit_obstacle"]:
```
mikeroberts3000, line 192: - mimic the `if args.benchmark ... else ...` structure in `generate_dataset.py`

examples/imitation_learning_openbot/generate_dataset.py
```diff
@@ -197,7 +160,7 @@
             action, policy_step_info = policy.step(obs)
```
mikeroberts3000, line 160: - rename `hit_obstacle = False` to `collect_data_for_episode = True` to improve readability

examples/imitation_learning_openbot/generate_dataset.py
```diff
@@ -217,16 +180,16 @@
...
+                time_data[i]        = time_stamp                                                                # current time stamp
+                frame_data[i]       = i                                                                         # current frame
 
             # debug
             if args.debug:
                 show_obs(
-                    obs, config.SIMULATION_CONTROLLER.OPENBOT_AGENT.OBSERVATION_COMPONENTS, config.SIMULATION_CONTROLLER.OPENBOT_AGENT.CAMERA.RENDER_PASSES)
+                    obs, config.SIMULATION_CONTROLLER.VEHICLE_AGENT.OBSERVATION_COMPONENTS, config.SIMULATION_CONTROLLER.VEHICLE_AGENT.CAMERA.RENDER_PASSES)
 
             # termination conditions
```
mikeroberts3000, line 194: It's fine for a policy to compute `waypoint_reached` because that is an implementation detail of the policy, and it is just used for debugging. But these policies should not be computing `goal_reached`, because that is an application-level responsibility.

```
            # update control action 
            action, policy_step_info = policy.step(obs)

            # send control action to the agent and collect observations
            obs, _, _, env_step_info = env.step(action={"set_duty_cycles": action})

            # compute goal reached condition
            # ...

            num_iterations = num_iterations + 1

            if args.debug:
                show_obs(obs, config.SIMULATION_CONTROLLER.VEHICLE_AGENT.OBSERVATION_COMPONENTS, config.SIMULATION_CONTROLLER.VEHICLE_AGENT.CAMERA.RENDER_PASSES)

            if not args.benchmark:
                obs_final_color = obs["camera.final_color"]
                assert len(obs_final_color.shape) == 3
                assert obs_final_color.shape[2] == 4
                obs_final_color = obs_final_color[:,:,[2,1,0,3]].copy() # note that spear.Env returns BGRA by default

                # save the collected rgb observations
                plt.imsave(os.path.realpath(os.path.join(image_dir, "%04d.jpg"%i)), obs_final_color)

                # During an episode, there is no guarantee that the agent reaches the predefined goal although its behavior is perfectly valid for training
                # purposes. In practice, it may for instance occur that the agent is not given enough time steps or control authority to move along the whole
                # trajectory. In this case, rather than considering the whole episode as a fail, one can consider the last position reached by the agent as
                # the new goal position. Doing so requires a recomputation of the compass observation, since the latter is goal dependant. Therefore, rather
                # than directly writing all the observations in a file iteration by iteration, we append these observations in a buffer, named "observation"
                # to later process them once the episode is completed. 
                control_data[i]     = action                                                                    # control_data: [ctrl_left, ctrl_right]
                state_data[i]       = np.concatenate((obs["location"], np.deg2rad(obs["rotation"])), axis=None) # state_data: [x, y, z, pitch, yaw, roll]
                waypoint_data[i]    = policy_step_info["current_waypoint"]                                      # current waypoint being tracked by the agent
                time_data[i]        = time_stamp                                                                # current time stamp
                frame_data[i]       = i                                                                         # current frame

            # termination conditions
            if env_step_info["task_step_info"]["hit_obstacle"]: 
                spear.log("    Collision detected, ending episode...")
                collect_data_for_episode = False
                break
            elif env_step_info["task_step_info"]["hit_goal"]: 
                spear.log("    Goal reached according env_step_info, ending episode...")
                break
            elif goal_reached: 
                spear.log("    Goal reached according to local distance calculation.")
                break
```

examples/getting_started/__init__.py
RachithP: We will need it in the `common` folder, but not here.

examples/getting_started/OpenBotEnv.py
RachithP: Yes, changed this.

cpp/unreal_plugins/SimulationController/Source/SimulationController/ImitationLearningTask.cpp
RachithP: I think I've now removed all the unwanted headers.

cpp/unreal_plugins/SimulationController/Source/SimulationController/ImitationLearningTask.cpp
RachithP: Updated this.

cpp/unreal_plugins/SimulationController/Source/SimulationController/ImitationLearningTask.cpp
RachithP: Updated this.

cpp/unreal_plugins/SimulationController/Source/SimulationController/ImitationLearningTask.cpp
RachithP: Updated this.

cpp/unreal_plugins/SimulationController/Source/SimulationController/ImitationLearningTask.h
RachithP: Updated this.

cpp/unreal_plugins/SimulationController/Source/SimulationController/NavMesh.h
RachithP: Updated this.

examples/getting_started/OpenBotEnv.py
RachithP: Updated this.

cpp/unreal_plugins/SimulationController/Source/SimulationController/NavMesh.cpp
RachithP: Updated this.

cpp/unreal_plugins/SimulationController/Source/SimulationController/NavMesh.cpp
RachithP: Why would `reference_points` be a multiple of 3? User should be able to pass any number of `reference_points` to this function.
Do you mean the length of each vector in `reference_points` should be 3?

cpp/unreal_plugins/SimulationController/Source/SimulationController/NavMesh.cpp
RachithP: Updated this.

cpp/unreal_plugins/SimulationController/Source/SimulationController/NavMesh.cpp
RachithP: Updated this.

cpp/unreal_plugins/SimulationController/Source/SimulationController/NavMesh.cpp
RachithP: Updated this.

cpp/unreal_plugins/SimulationController/Source/SimulationController/NavMesh.cpp
```diff
@@ -83,3 +90,72 @@ std::vector<uint8_t> NavMesh::getRandomPoints(int num_points)
...
+{
+    SP_ASSERT(start_points.size() == end_points.size());
+
+    std::vector<std::vector<uint8_t>> trajectories;
+
+    for (int i = 0; i < start_points.size(); ++i) {
+        SP_ASSERT(start_points.at(i).size() == 3);
+        SP_ASSERT(end_points.at(i).size() == 3);
+
+        FVector start_point_fvector{start_points.at(i).at(0), start_points.at(i).at(1), start_points.at(i).at(2)};
```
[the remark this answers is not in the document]
RachithP: Updated this.

cpp/unreal_plugins/SimulationController/Source/SimulationController/NavMesh.cpp
RachithP: Updated this.

examples/imitation_learning_openbot/generate_dataset.py
RachithP: Updated this.

cpp/unreal_plugins/SimulationController/Source/SimulationController/NavMesh.cpp
RachithP: I just saw your other comment. If we use `std::vector<double>` instead of `std::vector<std::vector<double>>`, we will need to assert the multiple of `3` condition.

cpp/unreal_plugins/SimulationController/Source/SimulationController/NavMesh.h
RachithP: 100% agree with the reasoning. For `getPaths()`, do you suggest we make both input and output `vector<vector>`? For simplicity, I think just the output being `vector<vector>>` should be sufficient.

cpp/unreal_plugins/SimulationController/Source/SimulationController/NavMesh.cpp
RachithP: Updated this.

cpp/unreal_plugins/SimulationController/Source/SimulationController/NavMesh.h
mikeroberts3000: I agree with you. The only reason for `getPaths()` to output a `vector` of `vectors` is because the output is "ragged", i.e., each path in the output can have a different number of points, and therefore there is no way to model the output as a dense 2D array. (If we wanted to output all the points in a single dense `vector` , we would also need to output a second `vector` with the number of points in each path, but this seems like too much effort. Since the input is not ragged, I think it is sufficient for the input to be a single vector.

examples/imitation_learning_openbot/policies.py
RachithP: Updated this.

examples/imitation_learning_openbot/generate_episodes.py
RachithP: Updated this.

examples/imitation_learning_openbot/generate_episodes.py
RachithP: Updated this.

examples/imitation_learning_openbot/generate_episodes.py
RachithP: Updated this.

examples/imitation_learning_openbot/generate_episodes.py
RachithP: Updated this.

examples/imitation_learning_openbot/generate_episodes.py
RachithP: Updated this. Also, corrected a typo in the indexing `candidate_best_indices  = np.argsort(candidate_scores)[:args.num_episodes_per_scene:-1]`. Instead used `candidate_best_indices  = np.argsort(candidate_scores)[::-1][:args.num_episodes_per_scene]`

examples/imitation_learning_openbot/generate_episodes.py
RachithP: Updated this.

examples/imitation_learning_openbot/generate_dataset.py
RachithP: Updated this.

cpp/unreal_projects/SpearSim/Config/DefaultGame.ini
```diff
@@ -8,6 +8,7 @@ ProjectID=51C548154BC1CB53E468FCA71A4A73EA
 
 [/Script/UnrealEd.ProjectPackagingSettings]
 bShareMaterialShaderCode=False ; we need this parameter to be False, otherwise Unreal will not find our materials
+bUseIoStore=False
```
mikeroberts3000, line 11:
```
bShareMaterialShaderCode=False ; we need this parameter to be False, otherwise Unreal will not find our materials
bUseIoStore=False              ; use Unreal's legacy asset loading system for compatibility with our existing approach in build_paks.py
```

examples/imitation_learning_openbot/generate_dataset.py
RachithP: Updated this

examples/imitation_learning_openbot/generate_dataset.py
RachithP: Updated this

examples/imitation_learning_openbot/generate_dataset.py
RachithP: Updated this

examples/imitation_learning_openbot/generate_dataset.py
RachithP: Updated this

examples/imitation_learning_openbot/generate_dataset.py
RachithP: Updated this

examples/imitation_learning_openbot/generate_dataset.py
RachithP: Updated this.

examples/imitation_learning_openbot/generate_dataset.py
RachithP: right, this is a bug, i.e., multiplying the `timestamp` value by `10000` will not give us time in `ns`. Updated the name to just `timestamp` as opposed to `timestamp[ns]`.

examples/imitation_learning_openbot/generate_dataset.py
RachithP: Updated this

examples/imitation_learning_openbot/generate_dataset.py
RachithP: Update this

examples/imitation_learning_openbot/generate_dataset.py
RachithP: Updated this

examples/imitation_learning_openbot/generate_dataset.py
RachithP: Updated this

examples/imitation_learning_openbot/generate_dataset.py
RachithP: Updated this.

examples/imitation_learning_openbot/policies.py
RachithP: Updated this

examples/imitation_learning_openbot/run_trained_policy.py
RachithP: I think now the overall code structure is as similar as possible to the code structure of `generate_dataset.py`

examples/imitation_learning_openbot/run_trained_policy.py
RachithP: Updated this

examples/imitation_learning_openbot/run_trained_policy.py
RachithP: Updated this

examples/imitation_learning_openbot/run_trained_policy.py
RachithP: Updated this

examples/imitation_learning_openbot/run_trained_policy.py
RachithP: Updated this

examples/imitation_learning_openbot/run_trained_policy.py
RachithP: Updated this

examples/imitation_learning_openbot/run_trained_policy.py
RachithP: Updated this

examples/imitation_learning_openbot/run_trained_policy.py
RachithP: Updated this

examples/imitation_learning_openbot/run_trained_policy.py
RachithP: I don't think we need a `else` condition here as we do not have to worry about data collection here.

examples/imitation_learning_openbot/run_trained_policy.py
RachithP: Updated this

cpp/unreal_projects/SpearSim/Config/DefaultGame.ini
RachithP: Updated this

examples/imitation_learning_openbot/utils.py
RachithP: Updated this

examples/imitation_learning_openbot/generate_dataset.py
```diff
@@ -11,12 +11,18 @@
...
-import spear
 import time
 
 from policies import *
+import spear
 from utils import *
 
+# hack to import OpenBotEnv from another example folder
+import sys
+SCRIPT_DIR = os.path.dirname(os.path.abspath(__file__))
```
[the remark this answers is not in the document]
RachithP: Updated this. We didn't have to call `dirname` twice.

examples/imitation_learning_openbot/policies.py
RachithP: Updated this

examples/imitation_learning_openbot/policies.py
RachithP: Updated this

examples/imitation_learning_openbot/policies.py
RachithP: Updated this

examples/imitation_learning_openbot/policies.py
RachithP: Updated this

examples/imitation_learning_openbot/policies.py
RachithP: Updated this

examples/imitation_learning_openbot/generate_dataset.py
RachithP: Updated this

examples/imitation_learning_openbot/generate_dataset.py
RachithP: Updated this

examples/imitation_learning_openbot/utils.py
RachithP: Updated the names to `location` instead of `poses` as we are only passing in `locations`.

examples/imitation_learning_openbot/utils.py
RachithP: Updated the names to location instead of poses as we are only passing in locations.

python/spear/env.py
RachithP: Updated this

examples/imitation_learning_openbot/generate_episodes.py
RachithP: Updated this

mikeroberts3000 (approved):
Looks great. Please review my changes double-check that everything is still working.

PR #216, opened 2023-07-19
Rachithp/release changes

RachithP:


examples/getting_started/user_config.yaml.example
```diff
@@ -35,8 +35,6 @@ SIMULATION_CONTROLLER:
   TASK: "NullTask"
   CUSTOM_UNREAL_CONSOLE_COMMANDS: []
 
-  SCENE_ID: "debug_0000"
```
mikeroberts3000, line 38: Let's keep this explicitly set to `apartment_0000`. Even though this is not strictly necessary, it will highlight to the readers of `getting_started` that this is how the scene can be changed.

examples/generate_image_dataset/poses.csv
```diff
@@ -1,101 +1,101 @@
 scene_id,location_x,location_y,location_z,rotation_pitch,rotation_yaw,rotation_roll
```
mikeroberts3000, line 1: Let's truncate to 10 poses, since this is the default behavior we define in `generate_poses.py`

mikeroberts3000 (approved):

PR #219, opened 2023-07-20
Rachithp/release changes

RachithP:


mikeroberts3000 (approved):
Looks good. I'm glad we don't need that ugly platform-specific if-statement in `setup.py`.

Issue #233, opened 2023-07-26
Running `generate_images.py` with `normal` render pass throws error


RachithP:
```
(spear) C:\github\spear\examples\generate_image_dataset>python generate_images.py
[SPEAR | env.py] Launching Unreal instance...
[SPEAR | env.py] Writing temp config file: C:\github\spear\examples\generate_image_dataset\tmp\config.yaml
[SPEAR | env.py] Launching executable with the following command-line arguments:
C:/github/spear/cpp/unreal_projects/SpearSim/Standalone-Shipping/Windows/SpearSim/Binaries/Win64/SpearSim-Win64-Shipping-Cmd.exe -windowed -resx=1024 -resy=1024 -graphicsadapter=0 -nosound -fileopenlog -stdout -fullstdoutlogoutput -nologtimes -renderoffscreen -config_file=C:\github\spear\examples\generate_image_dataset\tmp\config.yaml
[SPEAR | env.py] Launching executable with the following config values:
CORE_UTILS:
  WAIT_FOR_KEYBOARD_INPUT_DURING_INITIALIZATION: False
SIMULATION_CONTROLLER:
  AGENT: CameraAgent
  CAMERA_AGENT:
    ACTION_COMPONENTS: ['set_location', 'set_rotation']
    CAMERA:
      FOV: 90.0
      IMAGE_HEIGHT: 1024
      IMAGE_WIDTH: 1024
      RENDER_PASSES: ['normal']
    CAMERA_ACTOR_NAME: camera_actor
    OBSERVATION_COMPONENTS: ['camera']
    SPAWN_ACTOR_NAME:
    SPAWN_LOCATION_X: 0.0
    SPAWN_LOCATION_Y: 0.0
    SPAWN_LOCATION_Z: 200.0
    SPAWN_MODE: specify_pose
    SPAWN_ROTATION_PITCH: 0.0
    SPAWN_ROTATION_ROLL: 0.0
    SPAWN_ROTATION_YAW: 0.0
    STEP_INFO_COMPONENTS: []
  CAMERA_SENSOR:
    READ_SURFACE_DATA: True
    USE_SHARED_MEMORY: True
  CUSTOM_UNREAL_CONSOLE_COMMANDS: []
  IMITATION_LEARNING_TASK:
    AGENT_ACTOR_NAME:
    AGENT_SPAWN_OFFSET_LOCATION_X: 0.0
    AGENT_SPAWN_OFFSET_LOCATION_Y: 0.0
[...]

PR #236, opened 2023-07-28
Fix `SimulationController` plugin's assets names + `normal` render pass error

### LINKED ISSUE #232: Post processing materials present in `SimulationController` plugin's contents are incorrectly named
There are 3 post processing materials as part of `SimulationController` plugin's contents. It appears that the names were changed to `PPM_*` outside the editor, hence the change is not reflected in UE. If you open the Unreal Editor, you will see these materials named as `M_*` instead of `PPM_*`. This causes issues running `generate_image_dataset` example.

RachithP:
Fix #232, #233


examples/generate_image_dataset/generate_images.py
```diff
@@ -134,11 +134,11 @@ def single_step(self, action=None, get_observation=False):
 
                     # discard normals that aren't properly normalized, i.e., length of 1.0
                     discard_mask = np.logical_not(np.isclose(np.linalg.norm(obs_render_pass_vis, axis=2), 1.0, rtol=0.001, atol=0.001))
-                    obs_render_pass_vis = np.clip(0.0, 1.0, (obs_render_pass_vis + 1.0) / 2.0)
+                    obs_render_pass_vis = np.clip((obs_render_pass_vis + 1.0) / 2.0, 0.0, 1.0)
```
mikeroberts3000, line 137: This same bug (introduced by me, my bad 😅) also occurs a few lines above for the `depth` pass.

mikeroberts3000 (approved):
Please address my comments, but otherwise this looks great.

examples/generate_image_dataset/generate_images.py
RachithP: Updated.

PR #238, opened 2023-08-03
fix macOS codesign issue

RachithP:


tools/sign_executable/entitlements.plist
```diff
@@ -3,16 +3,6 @@
 <plist version="1.0">
 <dict>
 	<key>com.apple.application-identifier</key>
-	<string>org.embodiedaifoundation.spear</string>
```
mikeroberts3000, line 6: Nice

tools/sign_executable/sign_executable.py
```diff
@@ -43,103 +40,48 @@
...
+    sign_files = [
+        os.path.realpath(os.path.join(executable, "Contents", "UE", "Engine", "Binaries", "ThirdParty", "Intel", "TBB", "Mac", "libtbb.dylib")),
+        os.path.realpath(os.path.join(executable, "Contents", "UE", "Engine", "Binaries", "ThirdParty", "Intel", "TBB", "Mac", "libtbbmalloc.dylib")),
+        os.path.realpath(os.path.join(executable, "Contents", "UE", "Engine", "Binaries", "ThirdParty", "Ogg", "Mac", "libogg.dylib")),
+        os.path.realpath(os.path.join(executable, "Contents", "UE", "Engine", "Binaries", "ThirdParty", "Vorbis", "Mac", "libvorbis.dylib")),
+        os.path.realpath(os.path.join(executable, "Contents", "MacOS", os.path.splitext(executable_name)[0]))
+    ]
 
-        # send the zip file for notarization
+    assert os.path.exists(args.entitlements_file)
```
mikeroberts3000, line 54: Let's move this `assert` right below `parse_args()` because asserting earlier is better

tools/sign_executable/sign_executable.py
```diff
@@ -43,103 +40,48 @@
...
-        spear.log(f"Exceeded maximum wait time ({args.wait_time_seconds}s) for request UUID {request_uuid}. Please complete the rest of the procedure after notarization is complete.")
-        assert False
-            
+    subprocess.run(cmd, check=True)
+
+    # send the file for notarization
+    cmd = [
+        "xcrun", "notarytool", "submit", notarization_zip, "--apple-id", args.apple_id,
+        "--team-id", args.apple_teamid, "--password", args.apple_password, "--wait"
+    ]
```
mikeroberts3000, line 73: Use this style of square brackets when invoking `codesign` for consistency.

tools/sign_executable/sign_executable.py
```diff
@@ -43,103 +40,48 @@
     executable = os.path.realpath(os.path.join(args.output_dir, "Mac", "SpearSim-Mac-Shipping.app"))
     assert os.path.exists(executable)
 
-    if not args.request_uuid:
```
mikeroberts3000, line 46: Very nice reduction in complexity.

tools/sign_executable/sign_executable.py
```diff
@@ -43,103 +40,48 @@
...
+    spear.log(f"Executing: {' '.join(cmd)}")
+    ps = subprocess.Popen(cmd, stdout=subprocess.PIPE, text=True)
+    status = ""
+    for line in ps.stdout:
+        spear.log(f"{line}")
+        if "  status: " in line:
+            status = line.split("  status: ")[1].strip()
+    ps.wait()
+    ps.stdout.close()
+    assert status == "Accepted"
```
mikeroberts3000, line 84: I remember recently you did an extra debugging step where you ran a command-line tool, and you were able to examine a log from the notarization attempt. We should do this extra step after executing `notarytool`, write the log file to `temp_dir` and print it to the console, and then afterwards `assert` that status is accepted.

I'm not set up to run `sign_executable.py`, but if I were, it is a brittle enough procedure that I think I would always want to review the log, even during successful runs.

mikeroberts3000 (approved):

tools/sign_executable/sign_executable.py
RachithP: I was on the edge about adding this step. I think writing this log all the time is better because even when the status is successful, there might be some warnings logged in this log file, that is worth looking at (https://developer.apple.com/documentation/security/notarizing_macos_software_before_distribution/customizing_the_notarization_workflow)

RachithP: Added the command to download the log.

PR #254, opened 2023-09-20
SPEAR -> MuJoCo scene export

samarth-robo:
TODO 
- [x] move `environment.yml` to `setup.py`
- [x] improve `README.md`
- [x] reverse joint parent-child order, based on new SPEAR content guideline
- [ ] replace all Blender operations with `trimesh`


.gitignore
```diff
@@ -32,6 +32,7 @@ cpp/unreal_projects/SpearSim/SpearSimSource.pri
 cpp/unreal_projects/SpearSim/Saved
 cpp/unreal_projects/SpearSim/Standalone-Development
 cpp/unreal_projects/SpearSim/Standalone-Shipping
+cpp/unreal_projects/SpearSim/Temp
```
mikeroberts3000, line 35: What goes here?

.gitignore
```diff
@@ -40,6 +41,8 @@ cpp/unreal_projects/SpearSim/.vsconfig
 cpp/unreal_projects/SpearSim/.ignore
 cpp/unreal_projects/SpearSim/.kdev4
 
+.DS_Store
```
mikeroberts3000, line 44: Is this needed? If this was a result of some early prototyping, and we don't plan to generate any images in our base folder on an ongoing basis, this arguably isn't needed here.

.gitignore
```diff
@@ -54,10 +57,18 @@ examples/imitation_learning_openbot/tmp
 examples/imitation_learning_openbot/user_config.yaml
 examples/imitation_learning_openbot/videos
 
+examples/mujoco/tmp
```
mikeroberts3000, line 60: `examples/mujoco_interop`

.gitignore
```diff
@@ -54,10 +57,18 @@ examples/imitation_learning_openbot/tmp
...
+examples/mujoco/MUJOCO_LOG.TXT
+
 examples/open_loop_control_fetch/images
 examples/open_loop_control_fetch/tmp
 examples/open_loop_control_fetch/user_config.yaml
 
 python/spear_sim.egg-info
 
+tmp
```
mikeroberts3000, line 69: Why are these needed?
- `tmp`
- `tools/build`
- `tools/paks`

docs/getting_started.md
```diff
@@ -22,18 +22,22 @@ The next step is to install the `spear` Python package as follows.
...
+conda create --name spear-env python=3.10
 conda activate spear-env
 
 # install pip
 conda install -c anaconda pip
 
 # install msgpack-rpc-python separately from other Python dependencies so
 # we can use a specific commit from the msgpack-rpc-python GitHub repository
-pip install -e third_party/msgpack-rpc-python
+python -m pip install -e third_party/msgpack-rpc-python
```
mikeroberts3000, line 33: Why `python -m pip install` instead of `pip install`?

export/.gitignore
mikeroberts3000, line 1: These auto-generated .gitignore files are ugly and hard to read and edit. If we have things that we we're sure we want to be shared ignores for the whole team, let's put them in the global gitignore for simplicity (so they all live in one place) and also add them manually so we can group them consistently and logically.

mikeroberts3000, line 1: Rename `export` to `pipeline` or `data_pipeline` because it will soon contain scripts that import too. For example, once we are finished generating optimized collision geometry, we will want to import it back into Unreal. It makes sense to me that the script for doing so would be in the same folder as the other pipeline stages.

export/README.md
mikeroberts3000, line 1: Is this doc 100% up-to-date? We should polish it so it has a consistent tone and style with the other things in the `docs` folder, and rename it to `exporting_scenes_to_mujoco.md` for consistency with the other files in the `docs` folder.

export/README.md
```diff
@@ -0,0 +1,51 @@
+# MuJoCo Export Pipeline
+This code generates a MuJoCo replica of a SPEAR scene.
+
+## Usage
+All commands to be run from the `export` directory. Assume that the name of the scene is stored in the `SCENE_NAME`
+environment variable.
+- If you have not done so already, follow the SPEAR [Getting Started](../docs/getting_started.md) instructions, which
+will result in the `spear-env` conda environment.
+- Build V-HACD using the [readme instructions](v-hacd/README.md).
```
mikeroberts3000, line 9: If V-HACD is required to run this code, it should be added to the `third_party` folder and should be built automatically by `build_third_party_libs.py`

export/export_pipeline/__init__.py
mikeroberts3000, line 1: Do we want an `__init__` file here? We don't have one in the `tools` folder. Why is it desirable to have one here?

export/export_pipeline/coacd_thesh_calc.ipynb
mikeroberts3000, line 1: Are we using this notebook for anything? If we aren't using it, but we just want to check it in somewhere for posterity, then it should go in the `spear-internal/experimental` repo.

export/export_pipeline/compute_collision_representation.py
mikeroberts3000, line 1: - `compute_collision_geometry.py`
- `compute_gltf_scene.py`
- `compute_mujoco_scene.py`

export/export_pipeline/compute_collision_representation.py
```diff
@@ -0,0 +1,427 @@
...
+This script decomposes all static meshes into sets of convex regions.
+Decomposition method and parameters are based on the parameters chosen in UE, which are captured in 
+gltf_scene/actors_information.json.
+"""
+import argparse
+from collections import deque, defaultdict
+import copy
+import json
+import multiprocessing as mp
+if mp.current_process().name == 'MainProcess':
```
mikeroberts3000, line 11: - When is this false? `mp.current_process().name == 'MainProcess'`
- Find a better name for `utils.py` like `pipeline_utils.py` or `data_pipeline_utils.py`
- Our existing Python conventions: 4 spaces per indent, prefer `""` instead of `'` (consistency matters because it makes it easier to search for string literals)

- Move the weird `import` below the normal imports.

```
import argparse
from collections import deque, defaultdict
import copy
import json
import multiprocessing as mp
import numpy as np
import os
import shutil
import subprocess
import trimesh
import utils
import yaml

# coacd needs to be imported after bpy, otherwise terminate is called after throwing an std::bad_cast exception
if mp.current_process().name == "MainProcess":
    import bpy
import coacd
```

export/export_pipeline/compute_collision_representation.py
```diff
@@ -0,0 +1,427 @@
...
+import subprocess
+import trimesh
+import utils
+import yaml
+
+
+osp = os.path
+
+
+def decimation_lookup(poly_count):
```
mikeroberts3000, line 26: - Comment is more clear than the function name, rename to `get_adaptive_decimation_ratio(poly_count)` and delete comment.
- Implement this function using two arrays and the `bisect` function so the code is smaller and cleaner.

export/export_pipeline/compute_collision_representation.py
```diff
@@ -0,0 +1,427 @@
...
+    obj.select_set(True)
+    xyz = copy.deepcopy(obj.location)
+    xyz = [xyz.x, xyz.y, xyz.z]
+    quat = copy.deepcopy(obj.rotation_quaternion)
+    quat = [quat.w, quat.x, quat.y, quat.z]
+    obj.select_set(False)
+    return xyz, quat
+
+
+def get_relative_pose_and_center(obj, scale_factor=np.ones(3)):
```
mikeroberts3000, line 67: - Refactor into a read-only function (already implemented above) and a write-only function called `reset_pose(obj)` or something. The read-and-write responsibilities of this function are confusing.

export/export_pipeline/compute_collision_representation.py
```diff
@@ -0,0 +1,427 @@
...
+    xyz, quat = get_relative_pose(obj)
+    xyz = np.asarray(xyz) * scale_factor
+    obj.select_set(True)
+    obj.location = (0.0, 0.0, 0.0)
+    obj.rotation_quaternion = (1.0, 0.0, 0.0, 0.0)
+    obj.select_set(False)
+    return xyz.tolist(), quat
+
+
+def process_mesh(args: tuple) -> bool:
```
mikeroberts3000, line 80: - Don't pack `args` into a tuple because it makes the inputs of the function harder to read.

export/export_pipeline/compute_collision_representation.py
```diff
@@ -0,0 +1,427 @@
...
+    """
+    Decompose a mesh into convex regions according to hyperparameters.
+    Save the convex regions as well as a convex hull of the mesh in a directory structure.
+    args: a tuple containing (path to mesh, hyperparameters, rerun flag).
+    Returns a bool indicating success of the operation.
+    """
+    mesh_path, params, rerun = args
+    mesh_dir, mesh_filename = osp.split(mesh_path)
+    vhacd_args = params['collision_representation']['vhacd']
```
mikeroberts3000, line 89: - Use `yacs` with the `config.COLLISION_REPRESENTATION.VHACD` syntax for consistency with our other code.

export/export_pipeline/compute_collision_representation.py
```diff
@@ -0,0 +1,427 @@
...
+    coacd_args = params['collision_representation']['coacd']
+    acd_args   = params['collision_representation']['acd']
+
+    # Copy the obj file to the temporary directory.
+    decomposed = False
+    decompose_dir = osp.join(mesh_dir, params['common']['CONVEX_DECOMPOSITION_DIR'])
+    if rerun and osp.isdir(decompose_dir):
```
mikeroberts3000, line 96: I find this if statement confusing. If `rerun` is true (presumably because I want to "rerun" something), why does the if branch of this if statement do almost nothing? I think my confusion stems from the confusing variable name. If `rerun` is a flag to "re-run the export without running decomposition", then it should be named `skip_decomposition` or something. That would make the if statement less confusing, because then it would read as follows:

```
if args.skip_stuff:
    do_less_work()
else:
    do_more_work()
```

export/export_pipeline/compute_collision_representation.py
```diff
@@ -0,0 +1,427 @@
...
+        decomposed = True
+        os.remove(mesh_path)
+    else:
+        os.makedirs(decompose_dir, exist_ok=True)
+        mesh_path = shutil.move(mesh_path, decompose_dir)
```
mikeroberts3000, line 101: Why are we moving files around? Shouldn't we just generate the files in the desired locations so we don't need to move them later?

export/export_pipeline/compute_collision_representation.py
```diff
@@ -0,0 +1,427 @@
...
+        obj_name = os.path.splitext(mesh_filename)[0]
```
mikeroberts3000, line 102: Are we still using obj files?

export/export_pipeline/compute_collision_representation.py
```diff
@@ -0,0 +1,427 @@
...
+        
+        # load mesh and check that it is valid
+        mesh: trimesh.Trimesh = trimesh.load(mesh_path, force='mesh')
+        if mesh.is_empty:
+            volume = 0.0
+        else:
+            try:
+                volume = mesh.bounding_box_oriented.volume
+            except ValueError:
```
mikeroberts3000, line 111: Under what circumstances would `ValueError` happen?

export/export_pipeline/compute_collision_representation.py
```diff
@@ -0,0 +1,427 @@
...
+                volume = 0.0
+        if volume < 1e-6:
+            acd_args['decompose_method'] = 'skip'
```
mikeroberts3000, line 114: What is the difference between `skip` and `none`? Our convention is to treat config values as read-only unless there is a good reason not to.

export/export_pipeline/compute_collision_representation.py
mikeroberts3000: Actually, are these helper functions even used anywhere? If not, we should remove.

export/export_pipeline/compute_collision_representation.py
```diff
@@ -0,0 +1,427 @@
...
+                mcts_nodes=coacd_args['mcts_nodes'],
+                mcts_iterations=coacd_args['mcts_iterations'],
+                mcts_max_depth=coacd_args['mcts_max_depth'],
+                pca=coacd_args['pca'],
+                merge=coacd_args['merge'],
+                seed=coacd_args['seed'],
+                )
+
+            mesh_parts = []
+            for vs, fs in result:
```
mikeroberts3000, line 150: - These variable names are somewhat confusing because they're so terse.
- Convert `vs` and `fs` to numpy arrays so you can do this cleanup step with numpy fancy indexing, rather than resorting to slow ugly Python for-loops.
- Python List comprehensions are more elegant than raw for loops but will still be slow for large meshes.

export/export_pipeline/compute_collision_representation.py
```diff
@@ -0,0 +1,427 @@
...
+                final_fs = []
+                for f in fs:
+                    ff = frozenset(f)  # frozenset() also sorts
+                    if ff in fs_set:
+                        fs_set.remove(ff)
+                    else:
+                        fs_set.add(ff)
+                        final_fs.append(f)
+                mesh = trimesh.Trimesh(vs, final_fs)
+                if mesh.is_volume:
```
mikeroberts3000, line 164: Why do we use `mesh.is_volume` here but `mesh.is_empty` earlier?

export/export_pipeline/compute_collision_representation.py
```diff
@@ -0,0 +1,427 @@
...
+                fs_set = set()
+                final_fs = []
+                for f in fs:
+                    ff = frozenset(f)  # frozenset() also sorts
+                    if ff in fs_set:
+                        fs_set.remove(ff)
+                    else:
+                        fs_set.add(ff)
+                        final_fs.append(f)
+                mesh = trimesh.Trimesh(vs, final_fs)
```
mikeroberts3000, line 163: Rename `mesh` to `mesh_part` or `part` because it's going into an array called `mesh_parts`.

export/export_pipeline/compute_collision_representation.py
```diff
@@ -0,0 +1,427 @@
...
+                if mesh.is_volume:
+                    mesh_parts.append(mesh)
+
+            object_assembled = trimesh.Scene()
+            np.random.seed(0)
+            i = 0
+            for p in mesh_parts:
+                p.visual.vertex_colors[:, :3] = (np.random.rand(3) * 255).astype(np.uint8)
```
mikeroberts3000, line 171: Isn't this equivalent to the following simpler form:

```
p.visual.vertex_colors = ...
```

export/export_pipeline/compute_collision_representation.py
```diff
@@ -0,0 +1,427 @@
...
+                    if ff in fs_set:
+                        fs_set.remove(ff)
+                    else:
+                        fs_set.add(ff)
+                        final_fs.append(f)
+                mesh = trimesh.Trimesh(vs, final_fs)
+                if mesh.is_volume:
+                    mesh_parts.append(mesh)
+
+            object_assembled = trimesh.Scene()
```
mikeroberts3000, line 167: We call it an `object` here and a `mesh` earlier, but we're referring to the same underlying entity. I think it would be better if we made this more consistent.

export/export_pipeline/compute_collision_representation.py
```diff
@@ -0,0 +1,427 @@
...
+            acd_args['decompose_method'] = 'none'
+            print(f'{obj_name} is small, volume = {volume:.5f}, will not decompose')
+
+        # save convex hull
+        hull_mesh: trimesh.Trimesh = trimesh.convex.convex_hull(mesh)
+        hull_filename = osp.join(mesh_dir, 'convex_hull.stl')
+        hull_mesh.export(hull_filename)
+        print(f'Convex hull {hull_filename} saved.')
+        
+        if acd_args['decompose_method'] == 'coacd':
```
mikeroberts3000, line 126: I think it's worth it to break up each decompose method into it's own function.

export/export_pipeline/compute_collision_representation.py
```diff
@@ -0,0 +1,427 @@
...
+                    "-a",
+                    f"{int(not vhacd_args['disable_asyn'])}",
+                    "-l",
+                    f"{vhacd_args['min_edge_length']}",
+                    "-p",
+                    f"{int(vhacd_args['split_hull'])}",
+                ],
+                check=True,
+                cwd=mesh_dir,
+                stdout=subprocess.PIPE,
```
mikeroberts3000, line 225: I believe `PIPE` and `STDOUT` are the defaults and therefore do not need to be specified.

export/export_pipeline/compute_collision_representation.py
```diff
@@ -0,0 +1,427 @@
...
+                object_part.add_geometry(p)
+                filename = osp.join(decompose_dir, f"{i:03d}{params['common']['CONVEX_DECOMPOSITION_EXT']}")
+                object_part.export(filename)
+                object_assembled.add_geometry(p)
+                i+=1
+            if not object_assembled.is_empty:
+                object_assembled.export(osp.join(decompose_dir, 'assembled.obj'))
+                decomposed = True
+        elif acd_args['decompose_method'] == 'vhacd':
+            if not vhacd_args['enable']:
```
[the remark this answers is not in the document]
mikeroberts3000: Alternatively, if we were willing to remove V-HACD entirely, we could simplify this file significantly. My current thinking is that it is a good idea to leave V-HACD enabled for now as a debug/diagnostic tool, because it is the default algorithm used in Unreal, so it is arguably a useful baseline.

export/export_pipeline/compute_collision_representation.py
```diff
@@ -0,0 +1,427 @@
...
+                ],
+                check=True,
+                cwd=mesh_dir,
+                stdout=subprocess.PIPE,
+                stderr=subprocess.STDOUT,
+            )
+            output_str = ret.stdout.decode()
+            print(output_str)
+                    
+            if (ret.returncode != 0) or ('VolumeError' in output_str):
```
mikeroberts3000, line 231: Prefer asserts rather than error handling if statements.

export/export_pipeline/compute_collision_representation.py
```diff
@@ -0,0 +1,427 @@
...
+                print(f"V-HACD failed on {mesh_path}")
+            else: 
+                decomposed = True
+            shutil.move(osp.join(mesh_dir, vhacd_args['outputs'][1]), osp.join(decompose_dir, 'assembled.stl'))
```
mikeroberts3000, line 235: It seems inconsistent to me that some filenames (e.g., `assembled.stl`) are hard-coded in this file but some are specified via the config file.

mikeroberts3000, line 235: Moving files around as a post-process is a code smell. Why don't we just generate it in the desired location in the first place?

export/export_pipeline/compute_collision_representation.py
mikeroberts3000: Also, there are several variables at different scopes all named `mesh`. Our convention is to avoid this type of variable shadowing unless there is a good reason not to.

export/export_pipeline/compute_collision_representation.py
```diff
@@ -0,0 +1,427 @@
...
+            os.remove(osp.join(mesh_dir, vhacd_args['outputs'][0]))
+            os.remove(osp.join(mesh_dir, vhacd_args['outputs'][2]))
+            for src_filename in next(os.walk(decompose_dir))[-1]:
+                dst_filename = src_filename.replace(obj_name, '')
+                if dst_filename == '.stl':
+                    continue
+                shutil.move(osp.join(decompose_dir, src_filename), osp.join(decompose_dir, dst_filename))
+        elif acd_args['decompose_method'] == 'none':
```
mikeroberts3000, line 243: Why do we need `none` and `skip`?

export/export_pipeline/compute_collision_representation.py
```diff
@@ -0,0 +1,427 @@
...
+            out_filename = osp.join(decompose_dir, 'assembled.stl')
+            mesh.export(out_filename)
+            out_filename = osp.join(decompose_dir, '000.stl')
```
mikeroberts3000, line 246: Our convention in the Unreal Editor is to use four digits for everything. Let's use four digits here too for consistency.

export/export_pipeline/compute_collision_representation.py
```diff
@@ -0,0 +1,427 @@
...
+            if params['common']['CONVEX_DECOMPOSITION_EXT'] in mesh_path:
+                shutil.copy(mesh_path, out_filename)
+            else:
+                mesh.export(out_filename)
+            decomposed = True
+        elif acd_args['decompose_method'] == 'skip':
+            decomposed = False
+        else:
+            raise ValueError(f'{mesh_dir} decompose method {acd_args["decompose_method"]} is invalid')
```
mikeroberts3000, line 255: Our convention is to prefer asserts to exceptions.

export/export_pipeline/compute_collision_representation.py
```diff
@@ -0,0 +1,427 @@
...
+        
+        print(f"Removing {mesh_path}")
+        os.remove(mesh_path)
+        
+    if not decomposed:  # delete whole directory
```
mikeroberts3000, line 260: It can be a good idea to clear the output folder so there isn't any danger of mixing old and new files. For example, a script that generates a variable number of images might want to clear the output folder so you don't accidentally get images 0-10 from your most recent run, and then images 11-20 from your previous run.

But in this case, it would be better to do this at the beginning of a script, rather than at the end, because there might be intermediate output in the folder that would help to figure out why an error happened.

export/export_pipeline/compute_collision_representation.py
mikeroberts3000: - If this is a requirement imposed by `mp`, then write a wrapper called `process_mesh_mp` or something similar to unpack the args and call `process_mesh` with the unpacked args. This would keep the rest of the implementation separate from `mp` implementation details.

export/export_pipeline/compute_collision_representation.py
mikeroberts3000: Our convention is to use `dir` rather than `path` in variable names.

export/export_pipeline/compute_collision_representation.py
mikeroberts3000: - Why do we need `os.walk` here?

export/export_pipeline/compute_collision_representation.py
```diff
@@ -0,0 +1,427 @@
...
+        shutil.rmtree(mesh_dir)
+    return decomposed
+
+
+class CollisionRepresentationComputer(object):
```
mikeroberts3000, line 265: This object-oriented style does not seem to add much value.

export/export_pipeline/compute_collision_representation.py
```diff
@@ -0,0 +1,427 @@
...
+    def __init__(self, scene_path: str, n_workers: int, rerun: bool, include_objects: str) -> None:
+        self.n_workers = min(max(1, n_workers), mp.cpu_count()-1)
+        self.rerun = rerun
+        self.scene_path = osp.normpath(scene_path)
+        if include_objects is not None:
+            include_objects = include_objects.split(',')
+        self.input_gltf_filenames = []
+        with open(osp.join('export_pipeline', 'params.yaml'), 'r') as f:
```
mikeroberts3000, line 273: -Rename to `config.yaml` for consistency.

export/export_pipeline/compute_collision_representation.py
```diff
@@ -0,0 +1,427 @@
...
+            self.p = yaml.safe_load(f)
+        self.input_dir = osp.join(scene_path, self.p['common']['GLTF_SCENE_DIR_NAME'])
+        self.output_dir = osp.join(scene_path, self.p['common']['COLLISION_DIR_NAME'])
+        for filename in next(os.walk(self.input_dir))[2]:
```
mikeroberts3000, line 277: - Our convention is to implement this style of filesystem iteration using `glob`:

https://github.com/isl-org/spear/blob/main/tools/build_paks.py#L155

export/export_pipeline/compute_collision_representation.py
```diff
@@ -0,0 +1,427 @@
...
+            if not filename.endswith('.gltf'):
+                continue
+            if include_objects and not any([osp.splitext(filename)[0] == i for i in include_objects]):
+                continue
+            self.input_gltf_filenames.append(osp.join(self.input_dir, filename))
+
+
+    def run(self):
```
mikeroberts3000, line 285: I think this file could use one extra layer of hierarchy, breaking up the big logical chunks into named functions. `process_mesh` is good but it could be named better (e.g., `decompose_mesh`). This `run` function could be called `decimate_mesh` or something.

export/export_pipeline/compute_collision_representation.py
```diff
@@ -0,0 +1,427 @@
...
+        if not self.rerun:
+            shutil.rmtree(self.output_dir, ignore_errors=True)
+
+        bpy.ops.object.select_all(action='SELECT')
+        bpy.ops.object.delete()
+        
+        with open(osp.join(self.input_dir, 'actors_information.json'), 'r') as f:
```
mikeroberts3000, line 292: - Rename to `actors.json` because it is obvious that a json file contains information.

export/export_pipeline/compute_collision_representation.py
```diff
@@ -0,0 +1,427 @@
...
+    if not decomposed:  # delete whole directory
+        shutil.rmtree(mesh_dir)
+    return decomposed
+
+
+class CollisionRepresentationComputer(object):
+    def __init__(self, scene_path: str, n_workers: int, rerun: bool, include_objects: str) -> None:
+        self.n_workers = min(max(1, n_workers), mp.cpu_count()-1)
+        self.rerun = rerun
+        self.scene_path = osp.normpath(scene_path)
```
mikeroberts3000, line 269: Our convention is to use `realpath` over `normpath` unless there is a good reason not to.

export/export_pipeline/compute_collision_representation.py
```diff
@@ -0,0 +1,427 @@
...
+            obj = bpy.context.object
+            if obj.parent is not None:
+                continue
+            
+            object_name = obj.name.split('.')[0]
+            obj_dir = os.path.join(self.output_dir, object_name)
+            obj_info = actors_information[object_name]
+
+            joint_filename = input_gltf_filename.replace('.gltf', '_joints.json')
+            try:
```
mikeroberts3000, line 307: - Don't use exceptions for control flow.

export/export_pipeline/compute_collision_representation.py
```diff
@@ -0,0 +1,427 @@
...
+        process_mesh_args = []
+        for input_gltf_filename in self.input_gltf_filenames:
+            bpy.ops.import_scene.gltf(filepath=input_gltf_filename)
+            obj = bpy.context.object
+            if obj.parent is not None:
+                continue
+            
+            object_name = obj.name.split('.')[0]
+            obj_dir = os.path.join(self.output_dir, object_name)
+            obj_info = actors_information[object_name]
```
mikeroberts3000, line 304: - The variable names in this section are a bit too terse, and should be more more consistent:

```
# less consistent
obj_info = actors_information[object_name]

# more consistent
actor_desc = actor_descs[actor_name]
```

export/export_pipeline/compute_collision_representation.py
```diff
@@ -0,0 +1,427 @@
...
+            try:
+                with open(joint_filename, 'r') as f:
+                    joints_info = json.load(f)
+                joints_info = {j['name']: j for j in joints_info}
+            except FileNotFoundError:
+                joints_info = {}
+            
+            # traverse the tree, collect components info
+            geoms = []
+            components_info = {'static': {}, 'joints': {}}
```
mikeroberts3000, line 316: I agree with having top-level `static` and `joints` sections in our json file. But in our code, we don't need to create nested dictionaries that match our final json representation exactly when we are still manipulating and populating the dictionaries. It would be more readable if we had `static_mesh_components = {}` and `joint_components = {}` or something, and serialized them to json when we're finished manipulating them.

export/export_pipeline/compute_collision_representation.py
```diff
@@ -0,0 +1,427 @@
...
+            # we apply the scale factor to the transforms here and save the resulting transforms in components.json.
+            # geometry scaling is already applied by UE when exporting geometries to GLTF.
+            q = deque([(obj, obj_dir, np.ones(3), object_name), ])
+            while len(q):
+                o, dir, scale_factor, group_id = q.popleft()
+                name = o.name.split('.')[0]
+                xyz, quat = get_relative_pose_and_center(o, scale_factor)
+
+                if o.children or ((o.type == 'MESH') and (o.data.name != 'SM_Dummy')):
+                    if name in components_info['static']:
```
mikeroberts3000, line 327:
```
assert component_name not in static_mesh_components
```

export/export_pipeline/compute_collision_representation.py
```diff
@@ -0,0 +1,427 @@
...
+                        raise AssertionError(f'repeated static mesh {name}')
+                    components_info['static'][name] = {'pos': xyz, 'quat': quat}
+                    os.makedirs(dir, exist_ok=True)
+                    if o.children:  # StaticMeshComponent without any geometry (meant for grouping)
```
mikeroberts3000, line 331: A `StaticMeshComponent` can have non-trivial geometry and also have children. Is this comment up-to-date?

export/export_pipeline/compute_collision_representation.py
```diff
@@ -0,0 +1,427 @@
...
+            
+            # traverse the tree, collect components info
+            geoms = []
+            components_info = {'static': {}, 'joints': {}}
+            # each element is (bpy object, directory_path, scale_factor, group_id)
+            # we apply the scale factor to the transforms here and save the resulting transforms in components.json.
+            # geometry scaling is already applied by UE when exporting geometries to GLTF.
+            q = deque([(obj, obj_dir, np.ones(3), object_name), ])
+            while len(q):
+                o, dir, scale_factor, group_id = q.popleft()
```
mikeroberts3000, line 322: These variables are hard to understand. Because of the nested loop structure, it would help if the variable names hinted at the variable's scope. For example, I think the entries of `q` refer to Unreal components, so they should have `component` in the name:

```
component, component_dir, component_scale_factor, component_merge_id = components.popleft()
```

export/export_pipeline/compute_collision_representation.py
```diff
@@ -0,0 +1,427 @@
...
+                            else:
+                                group_id = None
+                            child_dir = osp.join(dir, child_name)
+                            q.append((child, child_dir, scale_factor*np.array(o.scale), group_id))
+                    else:  # StaticMeshComponent with geometry
+                        decompose_method = obj_info['geoms'][name]['decompose_method']
+                        geoms.append((dir, o, decompose_method, group_id))
+                elif (o.type == 'EMPTY') and not o.children:  # PhysicsConstraintComponent (i.e. joint)
+                    joint_name = utils.name_in_names(name, joints_info.keys())
+                    components_info['joints'][joint_name] = {
```
mikeroberts3000, line 346: Formatted inconsistently with line 329: `components_info['static'][name] = {'pos': xyz, 'quat': quat}`

export/export_pipeline/compute_collision_representation.py
```diff
@@ -0,0 +1,427 @@
...
+                    }
+
+            filename = osp.join(obj_dir, 'components.json')
+            with open(filename, 'w') as f:
+                json.dump(components_info, f, indent=4)
+
+            # group the leaves by group_id
+            # TODO(samarth) group name 
+            geom_groups = defaultdict(list)
+            for geom in geoms:
```
mikeroberts3000, line 358: It's hard to understand what this loop is doing because `geom` is a tuple. I think it would be better to use dicts instead of tuples for readability.

export/export_pipeline/compute_collision_representation.py
```diff
@@ -0,0 +1,427 @@
...
+                    )
+                bpy.ops.object.select_all(action='DESELECT')
+
+                params = copy.deepcopy(self.p)
+                params['collision_representation']['acd']['decompose_method'] = decompose_method
+                process_mesh_args.append((stl_filepath, params, self.rerun))
+
+        print("Convex solid decomposition...")
+        i = 0
+        while True:
```
mikeroberts3000, line 408: This while loop is ugly. Why not have a for-loop since you already know the number of iterates when the loop is executed. Additionally, this code would reduce CPU utilization in the case where some iterates are slow and some are fast. It seems like this could easily happen because some COACD invocations could be really complicated, and some are trivial.

You'd rather specify the full workload when calling `map`, to maximize the opportunities of each worker to steal other work if it finishes its own work quickly. Does `map` not allow you to do this?

Finally, any pipeline stage that uses `mp` to provide a single-process reference implementation that doesn't use `mp`.

export/export_pipeline/compute_collision_representation.py
```diff
@@ -0,0 +1,427 @@
...
+            with mp.Pool(self.n_workers) as p:
+                p.map(process_mesh, process_mesh_args[start_idx : end_idx])
+            i = i + 1
+            if end_idx >= len(process_mesh_args):
+                break
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    parser.add_argument('--scene_path', required=True, help='relative path to the scene folder')
```
mikeroberts3000, line 420: - Rename to `pipeline_dir` or `data_pipeline_dir`. I'm assuming that all pipeline output will go into some top-level folder.
- Rename to `num_parallel_workkers`, rename to `actors`.
- Include optional `scene_id` to run this pipeline stage for a single scene.
- Remove help strings for consistency with our other tools and because they always get out-of-date and we don't want to maintain them.

export/export_pipeline/compute_collision_representation.py
```diff
@@ -0,0 +1,427 @@
...
+    parser.add_argument('--rerun', action='store_true', help='re-run the export without running decomposition')
+    parser.add_argument('--objects', help='comma separated', default=None)
+    parser.add_argument('-n', type=int, help='number of parallel workers to use', default=mp.cpu_count()-1)
+    args = parser.parse_args()
+    
+    crc = CollisionRepresentationComputer(osp.expanduser(args.scene_path), args.n, args.rerun, args.objects)
```
mikeroberts3000, line 426: I think it would be a good idea to have a high-level function that acts like a table of contents and looks like this:

```
for actor_gltf in actor_gltfs:

    # compute a component desc for each component in the current per-actor GLTF
    merged_static_mesh_component_descs, physics_constraint_component_descs = compute_component_descs(...)

    # save component descs for downstream processing
    export_component_descs(merged_static_mesh_components, physics_constraint_components)

    # decimate each set of merged StaticMeshComponents, save the decimated results for downstream processing
    decimate_and_export_merged_static_mesh_components(merged_static_mesh_components)

    # compute convex decomposition for each set of merged StaticMeshComponents
    if args.use_single_threaded_reference_implementation:
        compute_and_export_decompositions(merged_static_mesh_components)

# compute convex decomposition for each set of merged StaticMeshComponents
if not args.use_single_threaded_reference_implementation:
    with mp.Pool(...) as pool:
        pool.map(...)
```

export/export_pipeline/compute_collision_representation.py
```diff
@@ -0,0 +1,427 @@
...
+            except FileNotFoundError:
+                joints_info = {}
+            
+            # traverse the tree, collect components info
+            geoms = []
+            components_info = {'static': {}, 'joints': {}}
+            # each element is (bpy object, directory_path, scale_factor, group_id)
+            # we apply the scale factor to the transforms here and save the resulting transforms in components.json.
+            # geometry scaling is already applied by UE when exporting geometries to GLTF.
+            q = deque([(obj, obj_dir, np.ones(3), object_name), ])
```
mikeroberts3000, line 320: Use a list and call `pop(0)` for consistency.

export/export_pipeline/compute_gltf_representation.py
mikeroberts3000, line 1: Move this into `cpp/unreal_projects/SpearSim/Content/Python` so it will be visible to Unreal by default. Create a lightweight script in `pipeline` (or `data_pipeline` or whatever, currently named `export_pipeline`) that runs UE and calls this script from the command-line.

export/export_pipeline/compute_gltf_representation.py
```diff
@@ -0,0 +1,270 @@
...
+from collections import namedtuple
+import json
+import os
+import time
+import unreal
+import yaml
+
+
+osp = os.path
+ss = unreal.get_editor_subsystem(unreal.UnrealEditorSubsystem)
```
mikeroberts3000, line 20: `editor_subsystem`, `editor_world`

export/export_pipeline/compute_gltf_representation.py
```diff
@@ -0,0 +1,270 @@
...
+world = ss.get_editor_world()
+
+
+TRANS_UE_TO_MUJOCO_SCALE = 1.0  # keep cm
```
mikeroberts3000, line 24: Since these are both 1.0, let's remove.

export/export_pipeline/compute_gltf_representation.py
```diff
@@ -0,0 +1,270 @@
...
+  options.set_editor_property('export_lights', False)
+  options.set_editor_property('export_uniform_scale', TRANS_UE_TO_MUJOCO_SCALE)
+  options.set_editor_property('export_unlit_materials', False)
+  options.set_editor_property('export_vertex_colors', False)
+  selected_actors = unreal.Set(unreal.Actor)
+  selected_actors.add(actor)
+  print(f'Writing actor {name} to {filename}...')
+  count = 0
+  N = 10
+  while not unreal.GLTFExporter.export_to_gltf(None, filename, options, selected_actors):
```
mikeroberts3000, line 50: - Why do we need this retry logic?
- A for loop would be cleaner.
- Assert rather than pass around error codes.

export/export_pipeline/compute_gltf_representation.py
```diff
@@ -0,0 +1,270 @@
...
+    print(f'Retry {count+1} / {N}')
+    time.sleep(0.5)
+    if count == 10:
+      break
+  else:
+    return True
+  return False
+
+
+PhysicsConstraint = namedtuple(
```
mikeroberts3000, line 61: Move these types immediately underneath imports for consistency.

export/export_pipeline/compute_gltf_representation.py
```diff
@@ -0,0 +1,270 @@
...
+  parser = argparse.ArgumentParser()
+  parser.add_argument('--scene_path', required=True)
+  parser.add_argument('--actors', default=None,
+                      help='Comma separate list of actor names - all others will be excluded if this option is used')
+  args = parser.parse_args()
+  filename = osp.join(osp.dirname(osp.abspath(__file__)), 'params.yaml')
+  with open(filename, 'r') as f:
+    params = yaml.safe_load(f)
+  # TODO(samarth): access params stuff by attributes
+  export_dir = osp.expanduser(osp.join(args.scene_path, params['common']['GLTF_SCENE_DIR_NAME']))
```
mikeroberts3000, line 78: `osp.realpath`

export/export_pipeline/compute_gltf_representation.py
```diff
@@ -0,0 +1,270 @@
...
+  os.makedirs(export_dir, exist_ok=True)
+  include_actors = args.actors.split(',') if args.actors else None
+  
+  filename = osp.join(export_dir, 'actors_information.json')
+  if osp.isfile(filename):
+    with open(filename, 'r') as f:
+      body_properties = json.load(f)
```
mikeroberts3000, line 85: We should be as consistent as possible with our names. Currently, the file is called `actors_information` but the variable is called `body_properties`. This is a code smell. I also don't really understand the logic here. Why are we loading a json file here? Isn't this part of the code the very beginning of our pipeline?

export/export_pipeline/compute_gltf_representation.py
```diff
@@ -0,0 +1,270 @@
...
+  else:
+    body_properties = {}
+  
+  # Iterate through the actors
+  for actor in unreal.GameplayStatics.get_all_actors_of_class(world, unreal.Actor):
+    if isinstance(actor, unreal.CameraActor) or isinstance(actor, unreal.SceneCapture2D) or \
```
mikeroberts3000, line 91: Instead of filtering by type and name, let's filter by path in the World Outliner and assert that we only encounter StaticMeshActors (not inherited) and Actors (not inherited).

export/export_pipeline/compute_gltf_representation.py
```diff
@@ -0,0 +1,270 @@
...
+    if 'HDRIBackdrop' in actor_name:
+      continue
+    if include_actors and (actor_name not in include_actors):
+      continue
+    
+    this_body_properties = {'geoms': {}, }
+    # check if actor has geometry
+    has_geometry = False
+    i = 0
+    for c in actor.get_components_by_class(unreal.StaticMeshComponent):
```
mikeroberts3000, line 104: `for component in ...`

export/export_pipeline/compute_gltf_representation.py
```diff
@@ -0,0 +1,270 @@
...
+      static_mesh = c.get_editor_property('static_mesh')
+      if static_mesh is None:
+        print(f'actor {actor_name} component {c.get_name()} does not have a static mesh')
+      elif static_mesh.get_name() != 'SM_Dummy':
+        # TODO(samarth) get MeshComponent elements from UE component
+        c_name = c.get_name()
```
mikeroberts3000, line 110: `component_name = ...`

export/export_pipeline/compute_gltf_representation.py
```diff
@@ -0,0 +1,270 @@
...
+        if c_name == 'StaticMeshComponent0':
+          c_name = actor_name
```
mikeroberts3000, line 112: I'm concerned about this special case. Does this make downstream processing simpler in any way?

export/export_pipeline/compute_gltf_representation.py
```diff
@@ -0,0 +1,270 @@
...
+    if isinstance(actor, unreal.CameraActor) or isinstance(actor, unreal.SceneCapture2D) or \
+      isinstance(actor, unreal.PointLight):
+      continue
+    actor_name = actor.get_actor_label()
+    if 'HDRIBackdrop' in actor_name:
+      continue
+    if include_actors and (actor_name not in include_actors):
+      continue
+    
+    this_body_properties = {'geoms': {}, }
```
mikeroberts3000, line 100:
```
actor_desc = {}
static_mesh_component_descs = {}
physics_constraint_component_descs = {}
```

export/export_pipeline/compute_gltf_representation.py
```diff
@@ -0,0 +1,270 @@
...
+    # check if actor has geometry
+    has_geometry = False
+    i = 0
+    for c in actor.get_components_by_class(unreal.StaticMeshComponent):
+      static_mesh = c.get_editor_property('static_mesh')
+      if static_mesh is None:
+        print(f'actor {actor_name} component {c.get_name()} does not have a static mesh')
```
mikeroberts3000, line 107: No need for this print statement. This is not an unexpected case.

export/export_pipeline/compute_gltf_representation.py
```diff
@@ -0,0 +1,270 @@
...
+      elif static_mesh.get_name() != 'SM_Dummy':
+        # TODO(samarth) get MeshComponent elements from UE component
+        c_name = c.get_name()
+        if c_name == 'StaticMeshComponent0':
+          c_name = actor_name
+        this_body_properties['geoms'][c_name] = MeshComponent('coacd', i)._asdict()
```
mikeroberts3000, line 113:
```
actor_desc.static_mesh_component_descs[component_name] = StaticMeshComponentDesc(decompose_method="coacd", merge_id=i)
```

export/export_pipeline/compute_collision_representation.py
```diff
@@ -0,0 +1,427 @@
...
+                    else:
+                        fs_set.add(ff)
+                        final_fs.append(f)
+                mesh = trimesh.Trimesh(vs, final_fs)
+                if mesh.is_volume:
+                    mesh_parts.append(mesh)
+
+            object_assembled = trimesh.Scene()
+            np.random.seed(0)
+            i = 0
```
mikeroberts3000, line 169: Use `enumerate`

export/export_pipeline/compute_gltf_representation.py
```diff
@@ -0,0 +1,270 @@
...
+    actor_name = actor.get_actor_label()
+    if 'HDRIBackdrop' in actor_name:
+      continue
+    if include_actors and (actor_name not in include_actors):
+      continue
+    
+    this_body_properties = {'geoms': {}, }
+    # check if actor has geometry
+    has_geometry = False
+    i = 0
```
mikeroberts3000, line 103: `i = 0 # can't use enumerate because we conditionally increment this counter`

export/export_pipeline/compute_gltf_representation.py
```diff
@@ -0,0 +1,270 @@
...
+      elif static_mesh.get_name() != 'SM_Dummy':
+        # TODO(samarth) get MeshComponent elements from UE component
+        c_name = c.get_name()
+        if c_name == 'StaticMeshComponent0':
+          c_name = actor_name
+        this_body_properties['geoms'][c_name] = MeshComponent('coacd', i)._asdict()
+        has_geometry = True
+        i += 1
+
+    if has_geometry:
```
mikeroberts3000, line 117: I think this loop over actors should be structured as follows.

```
    actor_desc.simulate_physics = ...

    for static_mesh_component in actor.get_components_by_class(unreal.StaticMeshComponent):
      ...
    
    if not actor_desc.has_non_trivial_geometry:
      continue

    for physics_constraint_component in actor.get_components_by_class(unreal.PhysicsConstraintComponent):
      ...

    actor_descs[actor_name] = actor_desc
    
    if actor_desc.has_non_trivial_geometry:
      export_gltf(...)
```

export/export_pipeline/compute_gltf_representation.py
```diff
@@ -0,0 +1,270 @@
...
+      export_meshes(actor, export_dir)
+    else:
+      print(f'Actor {actor_name}, does not have geometry')
+      continue
+    
+    if isinstance(actor, unreal.StaticMeshActor):
```
mikeroberts3000, line 123: - I don't think this logic is correct, `simulate_physics` is really a per-component property, so we probably should not try to collapse it to a per-actor property here.
- Let's call the property that we're saving `simulate_physics` because it's good to match the existing Unreal name unless we have a good reason to deviate.

export/export_pipeline/compute_gltf_representation.py
```diff
@@ -0,0 +1,270 @@
...
+      this_body_properties['moving'] = actor.get_editor_property('static_mesh_component').get_editor_property('body_instance').get_editor_property('simulate_physics')
+    else: # TODO(samarth) get the actual flag for the actuated body
+      this_body_properties['moving'] = False
+    this_body_properties['root_component_name'] = actor.get_editor_property('root_component').get_name()
+    body_properties[actor_name] = this_body_properties
+    
+    constraint_dicts = []
+    for component in actor.get_components_by_class(unreal.ActorComponent):
+      if not isinstance(component, unreal.PhysicsConstraintComponent):
```
mikeroberts3000, line 132: Why not iterate directly over `PhysicsConstraintComponents`?

export/export_pipeline/compute_gltf_representation.py
```diff
@@ -0,0 +1,270 @@
...
+        continue
+      
+      r = component.get_editor_property('relative_rotation')
+      rot = unreal.Rotator(r.roll, r.pitch, r.yaw)
+      
+      parent = str(component.get_editor_property('component_name2').get_editor_property('component_name'))
+      if (parent == 'DefaultSceneRoot') or (parent == 'StaticMeshComponent0'):
+        parent = actor_name
```
mikeroberts3000, line 140: Is this special case making our lives any easier?

export/export_pipeline/compute_gltf_representation.py
```diff
@@ -0,0 +1,270 @@
...
+      limit = 0.0
+      offset = 0.0
+      
+      linear_limit = constraint_profile.get_editor_property('linear_limit')
+      limit = linear_limit.get_editor_property('limit') * TRANS_UE_TO_MUJOCO_SCALE
+      
+      # prismatic joints
+      n_prismatic_dofs = 0
+      pc_args = [parent, child, ]
+      if linear_limit.get_editor_property('x_motion') != unreal.LinearConstraintMotion.LCM_LOCKED:
```
mikeroberts3000, line 156: I'm skeptical of the code complexity and repetition here. Why don't we just unconditionally export all the relevant joint information (`x_motion`, `y_motion`, `z_motion`, `swing1_motion`, `swing2_motion`, `twist_motion`, etc) and then attempt to construct the appropriate MuJoCo-centric concepts later.

export/export_pipeline/compute_gltf_representation.py
```diff
@@ -0,0 +1,270 @@
...
+      twist_limit = constraint_profile.get_editor_property('twist_limit')
+      pc_args = [parent, child, ]
+      if twist_limit.get_editor_property('twist_motion') != unreal.AngularConstraintMotion.ACM_LOCKED:
+        n_revolute_dofs += 1
+        if n_prismatic_dofs > 1:
+          raise ValueError(f'{actor_name} already has {n_prismatic_dofs} prismatic joints')
+        limit = twist_limit.get_editor_property('twist_limit_degrees') * ROT_UE_TO_MUJOCO_SCALE
+        offset = angular_offset.roll * ROT_UE_TO_MUJOCO_SCALE
+        # UE applied X rotations are right handed
+        # https://github.com/ethz-asl/unreal_airsim/blob/master/docs/coordinate_systems.md
```
mikeroberts3000, line 248: I'd rather refer to my description of Unreal rotations that I have previously shared in Discord (i.e., we should copy and paste to the top of this file). I'm skeptical about the correctness of the AirSim documentation.

export/export_pipeline/compute_gltf_representation.py
```diff
@@ -0,0 +1,270 @@
...
+          f'{component.get_name()}_x_revolute',
+          [-limit, limit],
+          offset,
+          [axis.x, axis.y, axis.z],
+          'revolute'
+        ])
+        pc = PhysicsConstraint(*pc_args)
+        constraint_dicts.append(pc._asdict())
+    
+    if len(constraint_dicts) > 0:
```
mikeroberts3000, line 261: If these json files are flat lists, I'd prefer reading and writing CSV files using pandas for simplicity and consistency.

export/export_pipeline/compute_collision_representation.py
samarth-robo: done. In the same spirit, changed `obj_name` to `mesh_name` in `process_mesh()`.

export/export_pipeline/compute_collision_representation.py
samarth-robo: Do you mean that this dictionary should not be formatted multi-line? If that's what you mean, I have corrected it.

export/export_pipeline/compute_collision_representation.py
samarth-robo: Agreed, changes made. I have made two namedtuples: `StaticMeshComponent` and `MergedStaticMeshComponents`. And used a dictionary `all_merged_smcs` which maps `decompose_dir` to `MergedStaticMeshComponents`. `decompose_dir` is the path to the convex decomposition directory. This path contains `merge_id` in it.

export/export_pipeline/compute_collision_representation.py
samarth-robo: Full workload consumes a large amount of memory. Please see the explanation I have written as a comment in the latest version. I have also replaced the while loop with a more succinct for loop.

export/export_pipeline/compute_collision_representation.py
samarth-robo: great idea, done!

export/export_pipeline/compute_gltf_representation.py
samarth-robo: we don't need the retry logic, removed it and asserted instead.

export/export_pipeline/compute_gltf_representation.py
samarth-robo: renamed `body_properties` to `actors_information`, and `this_body_properties` to `this_actor_information`. The JSON load at the beginning so that this script can append/modify to an existing `actors_information.json` if it is present. Let me know if you would like to overwrite it instead.

export/export_pipeline/compute_gltf_representation.py
samarth-robo: The component name is `StaticMeshComponent0` if it is a non-articulated actor. This is from my observation, but maybe it is not in the content guidelines that the name should be `StaticMeshComponent0` only.

I do this this makes downstream operations simpler. It serves the same purpose as renaming `DefaultSceneRoot` to `actor_name` for articulated actors.

export/export_pipeline/compute_gltf_representation.py
samarth-robo: done.

export/export_pipeline/compute_gltf_representation.py
samarth-robo: renamed `moving` to `simulate_physics`, need to discuss about the other point. It is mentioned as the unchecked item in README.md.

export/export_pipeline/compute_gltf_representation.py
samarth-robo: Please see this comment: https://github.com/isl-org/spear/pull/254#discussion_r1477598198.

mikeroberts3000:
Closing this PR because we have a functional MuJoCo interop demo checked into `main`.

PR #264, opened 2024-02-16
improved utils in Std and Unreal classes, camera enabled when paused

mikeroberts3000:


cpp/unreal_projects/SpearSim/Source/SpearSim/SpSpectatorPawn.cpp
```diff
@@ -0,0 +1,80 @@
...
+                spectator_pawn_movement_->MaxSpeed = spectator_pawn_movement_max_speed_ * 0.1;
+            } else {
+                // restore previous values
+                spectator_pawn_movement_->bIgnoreTimeDilation = spectator_pawn_movement_ignore_time_dilation_;
+                spectator_pawn_movement_->MaxSpeed = spectator_pawn_movement_max_speed_;
+            }
+            is_paused_ = is_paused;
+        }
+    #endif
+}
```
RachithP, line 80: Add a new line towards the EOF.

cpp/unreal_plugins/SpCore/Source/SpCore/Std.h
```diff
@@ -0,0 +1,233 @@
...
+    }
+
+    //
+    // Functions for safely reinterpreting contiguous value containers (e.g., std::initializer_list and std::vector)
+    //
+
+    template <typename TDest, CContiguousValueContainer TContiguousValueContainer>
+    static std::span<TDest> reinterpretAsSpanOf(TContiguousValueContainer& src)
+    {
+        size_t src_num_bytes = src.size() * sizeof(TContiguousValueContainer::value_type);
```
RachithP, line 202: Let's use `sizeof(typename TContiguousValueContainer::value_type)` instead of `sizeof(TContiguousValueContainer::value_type)` to be consistent with the rest of the code.

Also, isn't it necessary to use `typename` here? I wonder how it compiled without it 🤔

cpp/unreal_plugins/SimulationController/Source/SimulationController/StandaloneComponent.h
```diff
@@ -4,16 +4,17 @@
...
 
+#include <Components/ActorComponent.h>
 #include <Components/SceneComponent.h>
 #include <Engine/World.h>
 #include <GameFramework/Actor.h>
 
-#include "CoreUtils/Assert.h"
-#include "CoreUtils/Unreal.h"
+#include "SpCore/Assert.h"
+#include "SpCore/Unreal.h"
```
RachithP, line 15: I think it's good to add `// CActorComponent` alongside this include since it's not straightforward as to where `CActorComponent` is defined.

cpp/unreal_plugins/SimulationController/Source/SimulationController/StandaloneComponent.h
```diff
@@ -41,26 +40,5 @@ class StandaloneComponent
...
-    {
-        return Unreal::createSceneComponentOutsideOwnerConstructor<TComponent>(actor_, actor_, name);
-    }
-
-    template <>
-    TComponent* createComponent<false>(UObject* owner, AActor* parent, const std::string& name)
-    {
-        return Unreal::createComponentOutsideOwnerConstructor<TComponent>(actor_, name);
-    }
-
```
RachithP, line 64: Nice to see this go!

cpp/unreal_plugins/SpCore/Source/SpCore/Unreal.h
```diff
@@ -0,0 +1,423 @@
...
+
+#include <Components/ActorComponent.h>
+#include <Components/SceneComponent.h>
+#include <Containers/Array.h>        // TArray
+#include <Containers/StringConv.h>   // TCHAR_TO_UTF8, UTF8_TO_TCHAR
+#include <Containers/UnrealString.h> // FString::operator*
+#include <EngineUtils.h>             // TActorIterator
+#include <GameFramework/Actor.h>
+#include <HAL/Platform.h>            // TCHAR, TEXT
+#include <UObject/NameTypes.h>       // FName
```
RachithP, line 23: Let's align these comments.

cpp/unreal_plugins/SpCore/Source/SpCore/ArrayDesc.h
```diff
@@ -10,7 +10,7 @@
 #include <string>
```
RachithP, line 10: Let's align the header include comments in this file.

cpp/unreal_plugins/SimulationController/Source/SimulationController/RpcServer.h
```diff
@@ -156,14 +161,258 @@ struct FunctionWrapper<TReturn (*)(TArgs...)>
 
 } // namespace detail
 
-template <typename TFunctor>
-inline void RpcServer::bindAsync(const std::string& name, TFunctor&& functor)
+
```
RachithP, line 164: Do we still need the `wrapSync` function on Line 141?

RachithP (approved):
Few minor comments that need to be looked at. Everything else looks good.

cpp/unreal_plugins/SpCore/Source/SpCore/Std.h
mikeroberts3000: Great catch. This actually only compiles on Windows! I just verified on Compiler Explorer that `typename` is required on `clang`, just as you suspected.

cpp/unreal_plugins/SimulationController/Source/SimulationController/StandaloneComponent.h
mikeroberts3000: Great catch. 100% agreed.

cpp/unreal_plugins/SimulationController/Source/SimulationController/StandaloneComponent.h
mikeroberts3000: Agreed! I always thought this code was painful to look at. Haha.

cpp/unreal_plugins/SpCore/Source/SpCore/Unreal.h
mikeroberts3000: The rule I've been following so far is to align within each include block as far left as possible, but not to align across blocks.

cpp/unreal_plugins/SpCore/Source/SpCore/ArrayDesc.h
mikeroberts3000: The current alignment is consistent with the convention described above, since we put C headers in their own block, and the comments are aligned within each block. Do you disagree with the conventions?

cpp/unreal_plugins/SimulationController/Source/SimulationController/RpcServer.h
mikeroberts3000: No we do not. I didn't want to put more effort than necessary into this file though because you're in the middle of refactoring it. Maybe I shouldn't check in any changes to this file at all, because it'll cause downstream problems when you are merging your latest batch of changes.

mikeroberts3000:
Thank you for reviewing this in detail and finding some great things to fix! 😄

cpp/unreal_plugins/SimulationController/Source/SimulationController/RpcServer.h
```diff
@@ -156,14 +161,258 @@ struct FunctionWrapper<TReturn (*)(TArgs...)>
...
+template <typename TFunc, typename TReturn, typename... TArgs>
+concept CCallable =
+    requires(TFunc func, TArgs... args) {
+        { func(args...) } -> std::same_as<TReturn>;
+    };
+
+template <typename TBasicEntryPointBinder>
+concept CBasicEntryPointBinder =
+    requires(TBasicEntryPointBinder basic_entry_point_binder) {
+        { basic_entry_point_binder.template bind("", []() -> void {}) } -> std::same_as<void>;
```
RachithP, line 178: There's a typo here. This should be `basic_entry_point_binder.bind()`, and not `basic_entry_point_binder.template bind`.

cpp/unreal_plugins/SpCore/Source/SpCore/ArrayDesc.h
RachithP: Going with the convention mentioned [here] (https://github.com/isl-org/spear/pull/264#discussion_r1502140177), this is correct.

cpp/unreal_plugins/SimulationController/Source/SimulationController/RpcServer.h
RachithP: I agree that not modifying this file would be better as it'll help avoid unnecessary merging issues later on as you mentioned.

cpp/unreal_plugins/SimulationController/Source/SimulationController/RpcServer.h
RachithP: I wasn't aware of this, thanks for pointing it out!
I'm a bit curious now as it's still a bit unclear to me the requirement of `template` usage here. I understand that `bind` is a dependent name, dependent on the template `TBasicEntryPointBinder`. However, I'm trying to think of a scenario where the the compiler is confused while resolving `basic_entry_point_binder.bind(...)`. I understand the ambiguity while trying to resolve something like `basic_entry_point_binder.foo<...>(...)`, but in what scenario will there be an ambiguity in resolving what appears to be a simple function call 🤔

RachithP: Ah, my bad, figured it out. I didn't know `rpc::server`'s bind function was templated, here's the definition as found in `server.h` file from rpclib library.
```c++
    template <typename F> void bind(std::string const &name, F func) {
        disp_->bind(name, func);
    }
```

Since, `bind` is a templated function, we need to specify `template` keyword when we are using it as a dependent name.

mikeroberts3000: I'm also confused. I'm currently experimenting in Compiler Explorer, and and the `.template` syntax doesn't seem to be required by any compiler when calling the `bind(...)` function inside a template. But it isn't an error either. In other words, `gcc` and `clang` and `msvc` all consider it to be optional syntax when calling a function. I'm not sure why a compiler would be more strict inside a `requires` statement than in a templated function body.

mikeroberts3000: Actually never mind. This `.template` syntax isn't even required inside the `requires` clause on Windows! I'm sure I included it for a good reason at some point, because I was not even aware of this `.template` syntax before working on the new functionality in `RpcServer.h`, so I'm sure I included it to solve a specific error I encountered. But then my code must have evolved to the point where it was no longer necessary. I just removed it and everything compiles just fine.

RachithP: I'm trying to do the same, here's my code. I'm replicating the behavior of our `rpc::server` usage here.

```c++
#include <concepts>
#include <iostream>

// class that mimics rpc::server for the purpose of testing out usage of `Dummy.foo(...)` vs `Dummy.template foo(...)` only
class Dummy
{
public:
    template <typename T>
    void foo() {
        std::cout << "Dummy::foo" << std::endl;
    }
};

template <typename T>
concept CDot = requires(T t) {
    // { t.foo() } -> std::same_as<void>; // throws an error // error: no matching function for call to 'bar<B>(B&)'
    // { t.template foo() } -> std::same_as<void>; // throws an error // error: no matching function for call to 'bar<B>(B&)'
                                                   // note: the required expression 't.foo()' is invalid
    // { t.template foo<>() } -> std::same_as<void>; // throws an error // error: no matching function for call to 'bar<B>(B&)'
                                                   // note: the required expression 't.foo()' is invalid
    { t.template foo<int>() } -> std::same_as<void>; // this works
};

void bar(CDot auto x) {
    // x.foo(); // throws an error // error: no matching function for call to 'B::foo()'
    // x.template foo(); // throws an error // error: no matching function for call to 'B::foo()'
    // x.foo<>(); // throws a warning // warning: expected 'template' keyword before dependent template name [-Wmissing-template-keyword]
                  // and throws an error // expected primary-expression before '>' token, // expected primary-expression before ')' token
    // x.template foo<>(); // throws an error // error: no matching function for call to 'B::foo<>()'
    // x.foo<int>(); // throws a warning // warning: expected 'template' keyword before dependent template name [-Wmissing-template-keyword]
                     // throws an error // error: expected primary-expression before 'int', // error: expected ';' before 'int'
    x.template foo<int>(); // this works
    std::cout << "bar" << std::endl;
}

int main()
{
    Dummy b;
    bar<Dummy>(b);
    // bar(b); // this works too

    return 0;
}
```
`gcc 13.2` throws the above errors. `clang 17.0.1`, and `msvc v19.38` also agrees with `gcc`.

mikeroberts3000: My journey in this conversation 🤣 

image

mikeroberts3000: I don't think you've exactly replicated the correct experimental conditions in your code above. The `foo` function should be templated, so that part is correct. But you should try it with and without explicitly specifying the template parameter. When calling foo without the template parameter, the compiler needs to be able to infer the template parameter. I wouldn't expect this line to work under any circumstances,

```
    // { t.foo() } -> std::same_as<void>; // throws an error // error: no matching function for call to 'bar<B>(B&)'
```

because the template parameter can't be inferred. You should try the same experiment but with the `foo` function set up to allow the compiler to automatically infer the template parameter, e.g.,

```
template <typename T>
void foo(T t) {}
```

RachithP: You are right, this works
``` c++
#include <concepts>
#include <iostream>

// class that mimics rpc::server for the purpose of testing out usage of `Dummy.foo(...)` vs `Dummy.template foo(...)` only
class Dummy
{
public:
    template <typename T>
    void foo(T t) {
        std::cout << "Dummy::foo" << std::endl;
    }
};

template <typename T>
concept CDot = requires(T t) {
    { t.foo( []()->void{} ) } -> std::same_as<void>;                  // works
    // { t.template foo( []()->void{} ) } -> std::same_as<void>;      // works
    // { t.template foo<>( []()->void{} ) } -> std::same_as<void>;    // works
};

void bar(CDot auto x) {
    x.foo( []()->void{} );          // works
    // x.template foo( []()->void{} ); // msvc doesn't like this template keyword here
    // x.foo<>( []()->void{} );        // clang doesn't like not using template keyword here
    // x.template foo<>( []()->void{} );  // works
    std::cout << "bar" << std::endl;
}

int main()
{
    Dummy b;
    bar<Dummy>(b);
    bar(b);

    return 0;
}
```

RachithP: Conclusion:
- Looks like we need to use `template` keyword whenever there is a dependent template used with angled braces (`<>`). 
    - For e.g., in [this comment](https://github.com/isl-org/spear/pull/264#discussion_r1505069798), we can see that whenever `t.foo` is used in `t.foo<>` manner, we need the `template` keyword, or else `clang` (and `gcc`) will complain, although for some reason `msvc` doesn't.

- In order to be consistent across compilers, let's use `template` keyword.

mikeroberts3000: Great summary. Our bell curve meme needs a fourth section that is even further to the right! Haha.

cpp/unreal_projects/SpearSim/Source/SpearSim/SpSpectatorPawn.cpp
mikeroberts3000: Fixed.

cpp/unreal_plugins/SpCore/Source/SpCore/Std.h
mikeroberts3000: Fixed.

cpp/unreal_plugins/SimulationController/Source/SimulationController/StandaloneComponent.h
mikeroberts3000: Fixed. There were some other instances where this style of header comment was missing for our own headers, so I fixed those too.

cpp/unreal_plugins/SimulationController/Source/SimulationController/RpcServer.h
mikeroberts3000: Fixed. I removed all the changes from `RpcServer.h` except I needed to keep my adjustments to the `#includes` so the code would compile.

mikeroberts3000:
Note that I previously thought it would not be practical for our entry points to take `const` references as input arguments. I thought this because we often want to construct `std::span` objects from input `std::vector` objects, and I didn't think it was practical to construct a `std::span<T>` object from a `const std::vector<T>&` reference. But I realized this is actually pretty clean and easy. So we can keep the convention that our entry points take `const` references as input, and we don't need to use the `mutable` keyword in our new work queue implementation.

cpp/unreal_plugins/SpCore/Source/SpCore/Unreal.h
mikeroberts3000: Confirmed.

cpp/unreal_plugins/SpCore/Source/SpCore/ArrayDesc.h
mikeroberts3000: Confirmed.

PR #265, opened 2024-02-17
export every editor property on every actor, export scene geometry, compute kinematic tree representation, compute collision geometry, compute mujoco scene representation

mikeroberts3000:


mikeroberts3000:
This script outputs every property on every actor in the World Outliner in the following format. The output is quite verbose, so I'm only attaching the output for a couple of actors.

[outliner_short.json](https://github.com/isl-org/spear/files/14320277/outliner_short.json)

mikeroberts3000:
Added separate pipeline script to export scene geometry.

mikeroberts3000:
Added separate script to visualize the exported scene geometry.

mikeroberts3000:
Screenshot (84)

mikeroberts3000:
Screenshot (85)

python/setup.py
```diff
@@ -6,10 +6,24 @@
...
+        "mayavi==4.8.1",
+        "numpy==1.22.3",
+        "opencv-python==4.5.5.64",
+        "pandas==1.5.1",
+        "psutil==5.9.0",
+        "pyglet==1.5.0",
+        "tensorflow==2.13.0",
+        "trimesh[easy]==4.1.4",
+        "yacs==0.1.8",
+        "wxPython==4.2.1"
```
RachithP, line 27: Need to interchange these two.

mikeroberts3000: Good catch.

mikeroberts3000:
The pipeline now computes collision geometry in a separate pipeline stage.

mikeroberts3000:
Screenshot (89)

cpp/unreal_projects/SpearSim/Content/Python/pipeline_generate_unreal_metadata.py
```diff
@@ -0,0 +1,155 @@
...
+    # the CSV file by manually copying and pasting from the Unreal documentation.
+
+    # Get all Python attributes, and all editor properties from our CSV file.
+    uobject_class_names = [uobject.__class__.__name__] + [ base_class.__name__ for base_class in uobject.__class__.__bases__[::-1] ] # base-to-derived order
+    editor_property_names = set(dir(uobject))
+    for uobject_class_name in uobject_class_names:
+        editor_property_names = editor_property_names | set(df_editor_properties.loc[df_editor_properties["class"] == uobject_class_name]["editor_property"])
+
+    # Guess-and-check all editor properties.
+    editor_property_descs = dict()
```
mikeroberts3000, line 91: `dict()` -> `{}`

cpp/unreal_projects/SpearSim/Content/Python/pipeline_generate_unreal_metadata.py
```diff
@@ -0,0 +1,155 @@
...
+        return {
+            "class": editor_property.__class__.__name__,
+            "debug_info": {"str": str(editor_property)},
+            "editor_properties": get_editor_property_descs(editor_property)}
+
+    # Otherwise, if the editor property is an Unreal array, then recurse via get_editor_property_desc(...).
+    elif isinstance(editor_property, unreal.Array):
+        return [ get_editor_property_desc(editor_property_array_entry) for editor_property_array_entry in editor_property ]
+
+    # Otherwise, if the editor property value if serializable as JSON, then return the object,
```
mikeroberts3000, line 144: `is serializable`

pipeline/visualize_kinematic_trees.py
```diff
@@ -0,0 +1,156 @@
...
+        V_current_component = np.matrix(np.c_[mesh.vertices, np.ones(mesh.vertices.shape[0])]).T
+        V_world = M_world_from_current_component*V_current_component
+        assert np.allclose(V_world[3,:], 1.0)
+        mesh.vertices = V_world.T.A[:,0:3]
+
+        # Swap y and z coordinates to match the visual appearance of the Unreal editor.
+        if args.visual_parity_with_unreal:
+            mesh.vertices = mesh.vertices[:,[0,2,1]]
+
+        if args.color_mode == "unique_color_per_component":
```
mikeroberts3000, line 140: swap order of if statements

pipeline/visualize_unreal_geometry.py
```diff
@@ -0,0 +1,163 @@
...
+                    V_current_component = np.matrix(np.c_[mesh.vertices, np.ones(mesh.vertices.shape[0])]).T
+                    V_world = M_world_from_current_component*V_current_component
+                    assert np.allclose(V_world[3,:], 1.0)
+                    mesh.vertices = V_world.T.A[:,0:3]
+
+                    # Swap y and z coordinates to match the visual appearance of the Unreal editor.
+                    if args.visual_parity_with_unreal:
+                        mesh.vertices = mesh.vertices[:,[0,2,1]]
+
+                    if args.color_mode == "unique_color_per_component":
```
mikeroberts3000, line 147: swap order of if statements

pipeline/visualize_kinematic_trees.py
```diff
@@ -0,0 +1,156 @@
...
+z_axis_world = np.array([[0.0,   0.0,   100.0]])
+
+# Swap y and z coordinates to match the visual appearance of the Unreal editor.
+if args.visual_parity_with_unreal:
+    origin_world = origin_world[:,[0,2,1]]
+    x_axis_world = x_axis_world[:,[0,2,1]]
+    y_axis_world = y_axis_world[:,[0,2,1]]
+    z_axis_world = z_axis_world[:,[0,2,1]]
+
+np.random.seed(0)
```
mikeroberts3000, line 50: move up

python/setup.py
```diff
@@ -6,10 +6,23 @@
...
+        "matplotlib==3.6.2",
+        "mayavi==4.8.1",
+        "numpy==1.22.3",
+        "opencv-python==4.5.5.64",
+        "pandas==1.5.1",
+        "psutil==5.9.0",
+        "pyglet==1.5.0",
+        "tensorflow==2.13.0",
+        "trimesh[easy]==4.1.4",
+        "wxPython==4.2.1"
```
mikeroberts3000, line 27: vtk 9.2?

mikeroberts3000:
The pipeline now computes a functional MuJoCo scene in a separate pipeline stage.

mikeroberts3000:
Screenshot (91)

.gitignore
```diff
@@ -13,6 +13,7 @@ cpp/unreal_projects/SpearSim/Content/Scenes/kujiale_*
 cpp/unreal_projects/SpearSim/Content/Scenes/warehouse_*
 cpp/unreal_projects/SpearSim/Content/StarterContent
 cpp/unreal_projects/SpearSim/DerivedDataCache
+cpp/unreal_projects/SpearSim/enc_temp_folder
```
RachithP, line 16: What does `enc` mean here?

mikeroberts3000: It's a Visual Studio temp folder.

cpp/unreal_plugins/SpCore/Source/SpCore/Unreal.h
```diff
@@ -378,7 +372,22 @@ class SPCORE_API Unreal
...
+        if (include_stable_actor_name) {
+            AActor* actor = actor_component->GetOwner();
+            SP_ASSERT(actor);
+            component_name = getStableActorName(actor) + ":" + component_name;
+        }
+
+        return component_name;
+    }
+
+    static std::string getStableComponentName(const CSceneComponent auto* scene_component, bool include_stable_actor_name = false)
```
RachithP, line 390: I think it's a little confusing to have this function called `getStableComponentName` - when I first read this, it felt like I was getting the `StableName` of a component. Of course, we don't add `StableNameComponent` to a component, so it'd be wrong to think from that angle, but it's a bit confusing IMO. 

Why not just call this function `getComponentName`, and the boolean as `include_actor_name`? I think it'd be a little less confusing (i.e., less confusion between the meaning of `Stable` here vs `StableNameComponent`) with no `Stable` word.

python/setup.py
```diff
@@ -6,10 +6,25 @@
...
-    version="0.3.0",
+    version="0.5.0",
     author="",
     author_email="",
     description="",
     packages=["spear"],
-    install_requires=["ffmpeg-python==0.2.0", "gym==0.21.0", "matplotlib==3.6.2", "numpy==1.22.3", "opencv-python==4.5.5.64", "pandas==1.5.1", "psutil==5.9.0", "tensorflow==2.13.0", "yacs==0.1.8"]
-)
+    install_requires=[
+        "coacd==1.0.0"
```
RachithP, line 15: typo - there should be a `,` after `"coacd==1.0.0"`

python/setup.py
```diff
@@ -6,10 +6,25 @@
...
+        "mujoco==3.1.3",
+        "numpy==1.22.3",
+        "opencv-python==4.5.5.64",
+        "pandas==1.5.1",
+        "psutil==5.9.0",
+        "pyglet==1.5.0",
+        "tensorflow==2.13.0",
+        "trimesh[easy]==4.1.4",
+        "vtk==9.2.6",
+        "wxPython==4.2.1"
```
RachithP, line 29: typo - there should be a `,` after `"wxPython==4.2.1"`

cpp/unreal_plugins/SpCore/Source/SpCore/Unreal.h
mikeroberts3000: To me, the _stable name_ of an entity in Unreal means the unique string that we use throughout all SPEAR code (i.e., editor-only Python code, C++ code, MuJoCo code, etc) to refer to that entity. I like using the term "stable name" here to distinguish this concept from all of the other similar concepts in the Unreal Engine (`getName()`, `getActorLabel()`, etc).

When you call this function, you are getting the _stable name_ of the component, in the sense that you're getting the unique string that we use throughout all SPEAR code to refer to that component.

If we called this function `getComponentName`, I think it would be more confusing, because I don't know what underlying naming concept is being used, among all the following possibilities, all of which would be reasonable guesses:

```
actor->getName() + component->getName()
actor->getActorLabel() + component->getName()
root_component->getName() + child_component->getName() + grandchild_component->getName()
getStableActorName(actor) + component->getName()
...
```

I like calling this function `getStableComponentName` because a "stable name" has a specific meaning in SPEAR, so there is no guesswork to do to know what this function is going to return.

The fact that you need to add exactly one `StableNameComponent` to each actor in order to implement the `getStableActorName(actor)` function is an implementation detail. The fact that components don't need anything like that in order to implement the `getStableComponentName(component)` function is also an implementation detail. I don't think either implementation detail it should drive the naming conventions for these functions.

python/setup.py
mikeroberts3000: Good catch.

python/setup.py
mikeroberts3000: Good catch.

mikeroberts3000:
To set up the Python environment inside Unreal so it can run the scripts in the `Content/Python` folder, do the following steps from any command line, adjusting the paths as necessary for your machine.

```
H:\code\local\UnrealEngine\UE_5.2\Engine\Binaries\ThirdParty\Python3\Win64\python.exe -m pip install --disable-pip-version-check --no-warn-script-location -e F:\code\github\spear\third_party\msgpack-rpc-python
H:\code\local\UnrealEngine\UE_5.2\Engine\Binaries\ThirdParty\Python3\Win64\python.exe -m pip install --disable-pip-version-check --no-warn-script-location -e F:\code\github\spear\python
```

PR #269, opened 2024-03-18
access actors by a "stable name" that is consistent in the editor, in standalone builds, and in our import/export pipeline

mikeroberts3000:


cpp/unreal_plugins/SpCore/Source/SpCore/Unreal.h
```diff
@@ -23,6 +23,7 @@
 #include <UObject/NameTypes.h>       // FName
 
 #include "SpCore/Assert.h"
+#include "SpCore/StableNameComponent.h"
```
RachithP, line 26: `#include <type_traits> // std::is_base_of` is not required.

cpp/unreal_plugins/SpCore/Source/SpCore/StableNameComponent.h
```diff
@@ -0,0 +1,36 @@
...
+        void OnComponentCreated() override;
+        void PostLoad() override;
+    #endif
+
+    UPROPERTY(VisibleAnywhere, Category = "SPEAR", DisplayName = "Stable Name");
+    FString StableName;
+
+    // If we're in the editor, provide a public function to set StableName. This is useful, e.g., when the
+    // owning actor is renamed or moved in the World Outliner.
+#if WITH_EDITOR // defined in an auto-generated header
```
RachithP, line 33: Indentation is incorrect.

cpp/unreal_plugins/SpCore/Source/SpCore/StableNameComponent.cpp
```diff
@@ -0,0 +1,49 @@
...
+    {
+        AActor* actor = GetOwner();
+        SP_ASSERT(actor);
+        FName folder_path = actor->GetFolderPath();
+        if (folder_path.IsNone()) {
+            StableName = Unreal::toFString(Unreal::toStdString(actor->GetActorLabel()));
+            }
+        else {
+            StableName = Unreal::toFString(Unreal::toStdString(folder_path) + "/" + Unreal::toStdString(actor->GetActorLabel()));
+        }
```
RachithP, line 47: `if-else` condition formatting is incorrect.

cpp/unreal_plugins/SpCore/Source/SpCore/StableNameComponent.h
```diff
@@ -0,0 +1,36 @@
...
+        void update();
```
RachithP, line 34: Should this function's name be `updateStableName()` rather than `update()`, as the former conveys the purpose of this function more clearly.

RachithP (approved):
Some minor comments/feedback. Everything else looks great!

Also, question - What changes went into the `apartment_0000.umap`?

cpp/unreal_plugins/SpCore/Source/SpCore/StableNameComponent.h
mikeroberts3000: 9 times out of 10 I would agree with you. `update()` is usually too vague of a method name. But in this case I think it's fine, because any usage of this method would say something like `stable_name_component->update()`, which is the same level of information content as `stable_name_component->updateStableName()`, but is shorter.

mikeroberts3000:
I added a `StableNameComponent` to `SM_Prop_01`, which is the book object that you added to the scene to support the Fetch demo.

cpp/unreal_plugins/SpCore/Source/SpCore/StableNameComponent.cpp
mikeroberts3000: Good catch.

cpp/unreal_plugins/SpCore/Source/SpCore/StableNameComponent.h
mikeroberts3000: Good catch.

cpp/unreal_plugins/Vehicle/Source/Vehicle/VehiclePawn.cpp
```diff
@@ -138,8 +144,8 @@ AVehiclePawn::~AVehiclePawn()
 
     // Pawns don't need to be cleaned up explicitly.
 
-    SP_ASSERT(input_action_component_);
-    input_action_component_ = nullptr;
+    SP_ASSERT(user_input_component_);
+    user_input_component_ = nullptr;
 
     SP_ASSERT(MovementComponent);
     MovementComponent = nullptr;
```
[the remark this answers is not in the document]
mikeroberts3000: Good catch. As a matter of convenience, I made the change in my `pipeline` branch so it'll get corrected when the corresponding PR gets merged.

mikeroberts3000:
Screenshot (94)

PR #270, opened 2024-03-19
add ranges library support with UE 5.2 on Linux

RachithP:


cpp/unreal_projects/SpearSim/Source/SpTargetRules.Target.cs
```diff
@@ -36,7 +36,9 @@ public SpTargetRulesTarget(TargetInfo target) : base(target)
             // On macOS and Linux, we need to remap the paths of our symbolic links as we're compiling our executable, so the paths that get
             // written into the application's debug symbols aren't symbolic links. This is necessary to enable debugging in XCode and LLDB.
             bOverrideBuildEnvironment = true;
-            AdditionalCompilerArguments = "";
+            // The flag "-fexperimental-library" is required to enable support for ranges library. This is due to the clang-15.x compiler used in UE 5.2.
```
mikeroberts3000, line 39:
```
// The "-fexperimental-library" flag is required to enable support for std::ranges on Linux. This is because
// UE 5.2 builds with Clang 15 on Linux, but std::ranges are not fully supported in Clang 15 without this
// additional flag. The flag will not be necessary UE 5.3, which builds with Clang 16 on Linux.
```

cpp/unreal_plugins/SpCore/Source/SpCore/Std.h
```diff
@@ -187,7 +191,7 @@ class SPCORE_API Std
     static void insert(TKeyValueContainer& dest, const TKeyValueContainer& src)
     {
         std::vector<typename TKeyValueContainer::key_type> keys_intersection;
-        std::ranges::set_intersection(std::views::keys(dest), std::views::keys(src), std::back_inserter(keys_intersection));
+        std::ranges::set_intersection(keys(dest), keys(src), std::back_inserter(keys_intersection));
```
mikeroberts3000, line 194:
```
// TODO: revert back to using std::views after migrating to Unreal 5.3
//     std::ranges::set_intersection(std::views::keys(dest), std::views::keys(src), std::back_inserter(keys_intersection));
```

cpp/unreal_plugins/SpCore/Source/SpCore/Std.h
```diff
@@ -164,8 +167,9 @@ class SPCORE_API Std
     template <CKeyValueContainer TKeyValueContainer>
     static std::vector<typename TKeyValueContainer::key_type> keys(const TKeyValueContainer& key_value_container)
     {
-        auto view = std::views::keys(key_value_container);
-        return std::vector<typename TKeyValueContainer::key_type>(view.begin(), view.end());
+        std::vector<typename TKeyValueContainer::key_type> keys;
```
mikeroberts3000, line 170:
```
// TODO: revert back to using std::views after migrating to Unreal 5.3
//     auto view = std::views::keys(key_value_container);
//     return std::vector<typename TKeyValueContainer::key_type>(view.begin(), view.end());
```

mikeroberts3000 (approved):

PR #277, opened 2024-04-08
simplify user-facing python code + fix typos

RachithP:


cpp/unreal_plugins/SpCore/Source/SpCore/UnrealClassRegistrar.h
```diff
@@ -691,9 +683,44 @@ class SPCORE_API UnrealClassRegistrar
         s_get_static_class_registrar_.unregisterClass(class_name);
     }
 
+<<<<<<< HEAD
```
mikeroberts3000: There are merge issues here.

RachithP: Corrected it.

cpp/unreal_plugins/SpCore/Source/SpCore/Unreal.cpp
```diff
@@ -18,6 +18,7 @@
 #include <Containers/StringConv.h>   // TCHAR_TO_UTF8, UTF8_TO_TCHAR
 #include <Containers/UnrealString.h> // FString::operator*
 #include <Dom/JsonObject.h>
+#include <Dom/JsonValue.h>
```
mikeroberts3000, line 21: Great.

cpp/unreal_plugins/SpEngine/Source/SpEngine/Legacy/TickComponent.h
```diff
@@ -47,7 +47,7 @@ class UTickComponent : public UActorComponent
 
     void setTickFunc(const std::function<void(float, ELevelTick, FActorComponentTickFunction*)>& tick_func)
     {
-        tick_func_ = tick_func_;
+        tick_func_ = tick_func;
```
mikeroberts3000, line 50: Great.

examples/common/openbot_env_utils.py
mikeroberts3000, line 1: - `openbot_utils.py`
- shouldn't have an `Env` class, just have a standalone function that takes as input whatever state is needed to compute drive torques (I think you'll need to pass in the previous wheel speeds or something similar)

examples/imitation_learning_openbot/generate_dataset.py
```diff
@@ -41,17 +41,17 @@
     config = spear.get_config(user_config_files=[os.path.realpath(os.path.join(os.path.dirname(__file__), "user_config.yaml"))])
 
     config.defrost()
-    config.SIMULATION_CONTROLLER.TASK = "ImitationLearningTask"
-    config.SIMULATION_CONTROLLER.IMITATION_LEARNING_TASK.EPISODES_FILE = os.path.abspath(args.episodes_file)
+    config.SP_ENGINE.LEGACY_SERVICE.TASK = "ImitationLearningTask"
+    config.SP_ENGINE.LEGACY.IMITATION_LEARNING_TASK.EPISODES_FILE = os.path.abspath(args.episodes_file)
```
mikeroberts3000, line 45: Great.

examples/common/openbot_utils.py
mikeroberts3000, line 1: Implement a standalone function called `get_drive_torques()` in `openbot_utils.py`. This way, the `*_utils.py` files are all providing standalone functions that can be easily mixed-and-matched.

```
def get_drive_torques(self, duty_cycles, wheel_rotation_speeds, config):

    motor_velocity_constant = config.EXAMPLES.COMMON.OPENBOT_UTILS.MOTOR_VELOCITY_CONSTANT # Motor torque constant in [N.m/A]
    gear_ratio              = config.EXAMPLES.COMMON.OPENBOT_UTILS.GEAR_RATIO              # Gear ratio of the OpenBot motors
    motor_torque_constant   = 1.0 / motor_velocity_constant                                # Motor torque constant in [rad/s/V]

    battery_voltage         = config.EXAMPLES.COMMON.OPENBOT_UTILS.BATTERY_VOLTAGE         # Voltage of the battery powering the OpenBot [V]
    control_dead_zone       = config.EXAMPLES.COMMON.OPENBOT_UTILS.CONTROL_DEAD_ZONE       # Absolute duty cycle (in the ACTION_SCALE range) below which
                                                                                           # a command does not produce any torque on the vehicle

    motor_torque_max        = config.EXAMPLES.COMMON.OPENBOT_UTILS.MOTOR_TORQUE_MAX        # Motor maximal torque [N.m]
    electrical_resistance   = config.EXAMPLES.COMMON.OPENBOT_UTILS.ELECTRICAL_RESISTANCE   # Motor winding electrical resistance [Ohms]

    # Speed multiplier defined in the OpenBot to map a [-1,1] action to a suitable command to 
    # be processed by the low-level microcontroller. For more details, feel free to check the
    # "speedMultiplier" command in the OpenBot code.
    #     https://github.com/isl-org/OpenBot/blob/master/android/app/src/main/java/org/openbot/vehicle/Vehicle.java#L375
    action_scale = config.EXAMPLES.COMMON.OPENBOT_UTILS.ACTION_SCALE

    # Acquire the ground truth motor and wheel velocity for motor counter-electromotive force
    # computation purposes (or alternatively friction computation purposes).

    # The ground truth rotation speed of the motors in [rad/s]
    motor_speeds = gear_ratio*wheel_rotation_speeds

    # Compute the counter electromotive force using the motor torque constant
    counter_electromotive_forces = motor_torque_constant*motor_speeds # Expressed in [V]

    # The electrical current allowed to circulate in the motor is the result of the
    # difference between the applied voltage and the counter electromotive force
    motor_winding_currents = ((battery_voltage*duty_cycles) - counter_electromotive_forces)/electrical_resistance # Expressed in [A]

    # The torque is then obtained using the torque coefficient of the motor in [N.m]
    motor_torques = motor_torque_constant*motor_winding_currents

    # Motor torque is saturated to match the motor limits
    motor_torques = np.clip(motor_torques, -motor_torque_max, motor_torque_max)

    # The torque applied to the robot wheels is finally computed accounting for the gear ratio
    drive_torques = gear_ratio*motor_torques

    # Control dead zone at near-zero speed. This is a simplified but reliable way to deal with
    # the friction behavior observed on the real vehicle in the low-speed/low-duty-cycle regime.
    # For simplicity, we hardcode the threshold value 1e-5, but in future we should get this
    # value from the config system.
    drive_torques[np.logical_and(abs(motor_speeds) < 1e-5, abs(duty_cycles) <= control_dead_zone / action_scale)] = 0.0

    return drive_torques
```

Then implement an `OpenBotEnv` class in `openbot_env.py` that mixes-and-matches functionality from the `*_utils.py` files. Notice I'm mixing-and-matching the functionality from `openbot_utils` and `visualization_utils` internally here.

```
class OpenBotEnv(spear.Env):

    def __init__(self, config, instance):
    
        assert config.SP_ENGINE.LEGACY_SERVICE.AGENT == "VehicleAgent"
        assert "set_drive_torques" in config.SP_ENGINE.LEGACY.VEHICLE_AGENT.ACTION_COMPONENTS
        assert "wheel_rotation_speeds" in config.SP_ENGINE.LEGACY.VEHICLE_AGENT.OBSERVATION_COMPONENTS

        self._config = config
        self._wheel_rotation_speeds = None

        super().__init__(config, instance)

        # In this derived class, we are expecting different actions than the base spear.Env class. So
        # we need to override self.action_space. We need to do this after calling super().__init__(...),
        # because the base spear.Env class will set self.action_space to the lower-level actions it is
        # expecting internally.
        self.action_space = gym.spaces.Dict(spaces={"set_duty_cycles": gym.spaces.Box(low=-1, high=1, shape=(2,), dtype=np.float64)})
        self.observation_space = ...
    
    def reset(self, reset_info=None):
        obs = super().reset(reset_info=reset_info)
        assert "wheel_rotation_speeds" in obs.keys()
        self._wheel_rotation_speeds = obs["wheel_rotation_speeds"]
        return obs

    def _get_observation(self):
        obs = super()._get_observation()
        assert "wheel_rotation_speeds" in obs.keys()        
        self._wheel_rotation_speeds = obs["wheel_rotation_speeds"]
        if "camera.depth" in obs.keys():
            obs["camera.depth"] = visualization_utils.get_depth_image_for_visualization(obs["camera.depth"])
        return obs
        
    def _apply_action(self, action):
        assert "set_duty_cycles" in action.keys()
        assert action["set_duty_cycles"].shape[0] == 2
        assert self._wheel_rotation_speeds is not None
        drive_torques = openbot_utils.get_drive_torques(duty_cycles, self._wheel_rotation_speeds, self._config) 
        super()._apply_action(action={"set_drive_torques": drive_torques})
```

cpp/unreal_plugins/SpEngine/Source/SpEngine/LegacyService.h
```diff
@@ -39,6 +39,24 @@ class LegacyService {
         post_world_initialization_handle_ = FWorldDelegates::OnPostWorldInitialization.AddRaw(this, &LegacyService::postWorldInitializationHandler);
         world_cleanup_handle_ = FWorldDelegates::OnWorldCleanup.AddRaw(this, &LegacyService::worldCleanupHandler);
 
+        unreal_entry_point_binder->bindFuncNoUnreal("legacy_service", "get_random_points",
```
mikeroberts3000, line 42: This is not safe. `NoUnreal` functions can be called as soon as the RPC server is initialized, but `nav_mesh_` is only initialized in `LegacyService::worldBeginPlayHandler()`. (Also, the new convention I'd like to adhere to is calling these functions `Handlers` rather than `EventHandlers` for brevity.

examples/common/instance_utils.py
```diff
@@ -0,0 +1,42 @@
...
+# Copyright(c) 2022 Intel. Licensed under the MIT License <http://opensource.org/licenses/MIT>.
+#
+
+import spear
+
+def open_level(instance, scene_id, map_id=""):
+
+    def begin_tick(instance):
+        instance.engine_service.begin_tick()
+        instance.game_world_service.unpause_game()
```
mikeroberts3000, line 11: No need to pause and unpause the game here.

cpp/unreal_plugins/SpEngine/Source/SpEngine/LegacyService.h
RachithP: Nice catch! Updated the code to make navmesh related functions bind using `BindFuncUnreal` and associated usage on the python side.

examples/common/openbot_utils.py
RachithP: Made these changes.

cpp/unreal_plugins/SpCore/Source/SpCore/Log.cpp
```diff
@@ -17,8 +17,8 @@
 
 #include "SpCore/Assert.h"
 #include "SpCore/Boost.h"
-#include "SpCore/Unreal.h"
 #include "SpCore/Std.h"
+#include "SpCore/Unreal.h"
```
mikeroberts3000, line 21: Great.

cpp/unreal_plugins/SpCore/Source/SpCore/Log.h
```diff
@@ -10,8 +10,8 @@
 
 #include <CoreGlobals.h> // IsRunningCommandlet
 
-#include "SpCore/Std.h"
 #include "SpCore/Boost.h"
+#include "SpCore/Std.h"
```
mikeroberts3000, line 14: Great.

examples/common/instance_utils.py
mikeroberts3000, line 1:
```
    instance.engine_service.begin_tick()
    current_scene_id = instance.game_world_service.get_current_level()

    if current_scene_id != scene_id:
        desired_level_name = ""
        if scene_id != "":
            if map_id == "":
                map_id = scene_id
            else:
                map_id = map_id
            desired_level_name = "/Game/Scenes/" + scene_id + "/Maps/" + map_id

        spear.log("scene_id:           ", scene_id)
        spear.log("map_id:             ", map_id)
        spear.log("desired_level_name: ", desired_level_name)

        instance.game_world_service.open_level(desired_level_name)

    instance.engine_service.tick()
    instance.engine_service.end_tick()

    while current_scene_id != scene_id:
        instance.engine_service.begin_tick()
        current_scene_id = instance.game_world_service.get_current_level()
        instance.engine_service.tick()
        instance.engine_service.end_tick()
```

examples/common/openbot_env.py
```diff
@@ -38,8 +41,10 @@ def _get_observation(self):
         obs = super()._get_observation()
         assert "wheel_rotation_speeds" in obs.keys()
         self._wheel_rotation_speeds = obs["wheel_rotation_speeds"]
+        if "camera.depth" in obs.keys():
```
mikeroberts3000, line 44: You need to do the other modalities besides `depth` here.

examples/common/openbot_env.py
```diff
@@ -48,52 +53,7 @@ def _apply_action(self, action):
...
-
-        # The torque applied to the robot wheels is finally computed accounting for the gear ratio
-        drive_torques = gear_ratio * motor_torques
-
-        # Control dead zone at near-zero speed. This is a simplified but reliable way to deal with
-        # the friction behavior observed on the real vehicle in the low-speed/low-duty-cycle regime.
-        # For simplicity, we hardcode the threshold value 1e-5, but in future we should get this
-        # value from the config system.
-        drive_torques[np.logical_and(abs(motor_speeds) < 1e-5, abs(duty_cycles) <= control_dead_zone / action_scale)] = 0.0
+        drive_torques = get_drive_torques(duty_cycles, self._wheel_rotation_speeds, self._config)
```
mikeroberts3000, line 56:
```
        assert "set_duty_cycles" in action.keys()
        assert action["set_duty_cycles"].shape[0] == 2
        assert self._wheel_rotation_speeds is not None
        duty_cycles = np.array([action["set_duty_cycles"][0], action["set_duty_cycles"][1], action["set_duty_cycles"][0], action["set_duty_cycles"][1]], dtype=np.float64)
        drive_torques = get_drive_torques(duty_cycles, self._wheel_rotation_speeds, self._config)
        super()._apply_action(action={"set_drive_torques": drive_torques})
```

examples/common/openbot_utils.py
```diff
@@ -0,0 +1,54 @@
...
+#
+
+import numpy as np
+
+
+def get_drive_torques(duty_cycles, wheel_rotation_speeds, config):
+
+    motor_velocity_constant = config.EXAMPLES.COMMON.OPENBOT_UTILS.MOTOR_VELOCITY_CONSTANT # Motor torque constant in [N.m/A]
+    gear_ratio              = config.EXAMPLES.COMMON.OPENBOT_UTILS.GEAR_RATIO              # Gear ratio of the OpenBot motors
+    motor_torque_constant   = 1.0 / motor_velocity_constant              # Motor torque constant in [rad/s/V]
```
mikeroberts3000, line 12: Fix spacing.

examples/common/visualization_utils.py
```diff
@@ -0,0 +1,72 @@
+#
+# Copyright(c) 2022 Intel. Licensed under the MIT License <http://opensource.org/licenses/MIT>.
+#
+
+import numpy as np
+
+def get_depth_image_for_visualization(observation):
```
mikeroberts3000, line 7: Rename `observation` to `image` so it matches the function name.

examples/common/visualization_utils.py
```diff
@@ -0,0 +1,72 @@
...
+    return modified_observation
+
+def get_segmentation_image_for_visualization(observation):
+
+    assert len(observation.shape) == 3  # width, height, #channels
+    assert observation.shape[2] == 4    # RGBA or BGRA
+
+    return observation.copy()[:,:,[2,1,0]] # segmentation is returned as BGRA
+
+def get_observation_components_modified_for_visualization(observation, observation_components_to_modify):
```
mikeroberts3000, line 49: Let's remove this function. I like the other ones better because they're easier to use correctly.

examples/generate_image_dataset/generate_images.py
mikeroberts3000, line 1: Let's re-order the args of all `Env` constructors as `(instance, config, ...)` instead of `(config, instance, ...)`.

examples/generate_image_dataset/generate_poses.py
mikeroberts3000, line 1: Great. Very clean.

examples/getting_started/user_config.yaml.example
```diff
@@ -2,21 +2,23 @@
...
 
-OPENBOT_ENV:
-    # Motor torque: 1200 gf.cm (gram force centimeter) == 0.1177 N.m
-    # Gear ratio: 50
-    # Max wheel torque: 5.88399 N.m
-    # https://www.conrad.de/de/p/joy-it-com-motor01-getriebemotor-gelb-schwarz-passend-fuer-einplatinen-computer-arduino-banana-pi-cubieboard-raspbe-1573543.html
-    
-    MOTOR_VELOCITY_CONSTANT: 65.89 # Velocity constant of the OpenBot motors
-    GEAR_RATIO: 50.0               # Gear ratio of the OpenBot motors
+EXAMPLES:
```
mikeroberts3000, line 5: Let's move this to a `default_config.yaml` in the `common` folder and pass it the path when calling `get_config(...)`. We went to all the trouble of defining the class in `common` so we wouldn't need to repeat ourselves in each example. It's weird to go to all of that trouble and still need to repeat ourselves in each `user_config.yaml.example` file.

examples/imitation_learning_openbot/generate_episodes.py
```diff
@@ -12,6 +12,38 @@
...
 from utils import plot_paths
 
+# import open_level functionality from common folder
+common_dir = os.path.realpath(os.path.join(os.path.dirname(__file__), ".."))
+import sys
+sys.path.append(common_dir)
+from common.instance_utils import open_level
+
+
+class NavMesh:
```
mikeroberts3000, line 22: Move `NavMesh` to `common` because you're referring to it in multiple examples. And if you were only referring to it in this one example, a better place to put it would be in `utils.py`, not in `generate_episodes.py`.

python/spear/env.py
```diff
@@ -31,71 +30,30 @@ def __init__(self, config, instance):
...
-        while current_scene_id != scene_id:
-            Env.begin_tick(instance)
-            current_scene_id = instance.game_world_service.get_current_level()
-            Env.tick(instance)
-            Env.end_tick(instance)
-
     def step(self, action):
 
-        Env.begin_tick(self._instance)
+        self.begin_tick()
```
mikeroberts3000, line 35: Much better.

mikeroberts3000 (approved):

mikeroberts3000:
Please address outstanding comments.

examples/imitation_learning_openbot/generate_episodes.py
RachithP: Makes sense. Moved it to `common`.

mikeroberts3000:
@RachithP

- Rename `nav_mesh_env.py` to `nav_mesh.py` to match the name of the class in the file. In my previous comment, I didn't mean that we must name all files that mix-and-match `utils` functionality as `*_env.py`. I just meant that we should cleanly separate standalone functionality in `_utils.py` files and higher-level functionality in other files.
- Move functionality into `cpp` files that doesn't need to be in the service headers. Our current convention is, if something doesn't need to be in a header, it shouldn't be. We can make exceptions for extremely simple one-liner functions, if putting them in a header would make it so we don't need a `cpp` file. But the service implementations don't meet that criteria.

RachithP:
> @RachithP
> 
> * Rename `navmesh_env.py` to `navmesh.py` to match the name of the class in the file. In my previous comment, I didn't mean that we must name all files that provide higher-level functionality as `*_env.py`. I just meant that we should cleanly separate standalone functions in `_utils.py` files, and higher-level functionality in other files.

Made these changes in https://github.com/isl-org/spear/pull/279.

> * Move all service functionality that doesn't need to be in a header into `cpp` files. If something doesn't need to be in a header, it shouldn't be. We can make exceptions for extremely simple one-liner functions, if putting them in a header would make it so we don't need a `cpp` file at all. But the service implementations don't meet that criteria.

Since our `EngineService` class is templated, we can't move it to a `cpp` file. However, only constructors of other service classes are templated, so we can move their member functions to  a `cpp` file. Made these changes in https://github.com/isl-org/spear/pull/279 .

mikeroberts3000:
> Note that to maintain symmetry with the constructors, I kept the destructors in the header file.

Great choice. It's nice to see the constructor next to the destructor, even if we could move the destructor to a cpp file.

PR #281, opened 2024-04-22
update /Content/Kujiale

rwang15:
* cleanup naming convension for blueprint actors in /Content/Kujiale
* add base material for kujiale material v2 (used for kujiale_0026 and beyond)


cpp/unreal_projects/SpearSim/Content/Kujiale/BluePrints/BP_Ceiling_Light_00.uasset
mikeroberts3000, line 1: - Change capitalization from `BluePrints` to `Blueprints` for consistency with the Unreal documentation. I've commented on this same issue previously.
- Use consistent number of digits. It is confusing to have `BP_Ceiling_Light_00` and `BP_Ceiling_Light_0000` in the same folder. Let's use four digits for consistency with the rest of our naming conventions in the Kujiale data.

cpp/unreal_projects/SpearSim/Content/Kujiale/Materials/MtlFns/MF_CG_2d_TexOffset.uasset
mikeroberts3000, line 1: - Rename from `MtlFns` to `MaterialFunctions` for consistency with our existing folders. We call our existing folders `Materials` and `MaterialInstances`, not `Mtls` and `MtlInsts`. What does `CG` mean in these names?

mikeroberts3000:
This PR looks good. That being said, I'd like to review one more time before approving. (Although my recommended changes are minor and cosmetic, the last time I approved a PR with only minor cosmetic changes remaining, not all of those changes ended up in `main`. So I'd like to review one more time.)

rwang15:
* CG in MaterialFunctions - all funcions with CG are cherry picked from another plugin, and functions we added.
* renanming base material is a super expensive action where you need to refactor all of the material instance ever uploaded, and the change need to be done very carefully to not break any reference in Unreal. It could get very messy easily. We might need to find a good way to do if if we really want to rename them. 
* Renaming BP_Ceiling_Light_00 to BP_Ceiling_Light_0000 (also renaming the folder name) require changing the reference for all umaps at the same time to avoid broken reference. Im working on it but have some issue with unreal.

mikeroberts3000:
> * CG in MaterialFunctions
>     * all funcions with CG are cherry picked from another plugin, and functions we added.

That makes sense.

> * renanming base material 
>     * it is a super expensive action where you need to refactor all of the material instance ever uploaded, and the change need to be done very carefully to not break any reference in Unreal. It could get very messy easily. We might need to find a good way to do if if we really want to rename them.
> * Renaming BP_Ceiling_Light_00 to BP_Ceiling_Light_0000 (also renaming the folder name) 
>     * it also requires changing the reference for all umaps at the same time to avoid broken reference. Fixing redirectory at the same time for all maps would cause Unreal to crash because it load too many asset at the same time.  Im working on it but have some issue with unreal.

Fair enough. We can wait to pursue these improved naming conventions until we have a better sense of how to do it cleanly and safely, and how much effort it will be. I agree that it is not a good idea to attempt to load all Kujiale maps simultaneously and try to get the Unreal Editor to do it for us. We will need to write code that can do this for us automatically in a way that guarantees correctness, and can process one scene at a time.

That being said, I don't think we should wait indefinitely. If we can't afford to put effort into this now (or at least soon), how are we ever going to afford to put effort into this when we have an order of magnitude more scenes to deal with?

And it is clearly a confusing convention to have, e.g., `BP_Ceiling_Light_00` to `BP_Ceiling_Light_0000` in the same folder. If we too liberally apply the argument that the existing conventions are too hard to change, it will paralyze us, and make it so we never improve things. We could have applied this same argument to almost all of the improvements to our naming conventions we have made over the last 6-12 months. But if we did that, our data would be an incomprehensible mess and would be very difficult to work with. I like how the legibility of our data has improved over time, and I'm glad we invested effort into improving it, even though it was took a lot of work. I don't want to slide backwards to how our data used to be, because we have decided that our conventions are too hard to change.

mikeroberts3000 (approved):
I take your point that some of the naming improvements I have suggested are too heavyweight to consider for now. Please please make whatever improvements you think are appropriate for now, and if you think something is too heavyweight, then make a note of it here and we can revisit later. I would like to revisit this issue of naming conventions after the ICRA demo.

rwang15:
* base material - use old naming convension
* blueprint actors - new naming convension

The references in umaps are all fixed except kujiale_0005 (there is some error opening the level). I fixed the redirectory per blueprint with the option fixup(keep redirectory).

PR #285, opened 2024-05-17
Linux compile issue fix

RachithP:


mikeroberts3000 (approved):
We should treat `uint64` as an Unreal type, i.e., we should use it when we have a good reason to, like passing a value to an Unreal function that expects a `uint64`. Otherwise, we should use `uint64_t` in our interfaces, because it is a more standard type, unless we have a good reason not to. This PR brings our code into better alignment with this proposed convention, and apparently it fixes some platform-specific compilation errors. This is great.

RachithP:
Agreed, Unreal does this https://github.com/EpicGames/UnrealEngine/blob/463443057fb97f1af0d2951705324ce8818d2a55/Engine/Source/Runtime/Core/Public/HAL/Platform.h#L951C25-L951C31

PR #287, opened 2024-05-21
upgrade build tools related scripts to scale up 

RachithP:
- Update python dependencies

### DIFF
--- a/tools/build_executable.py
+++ b/tools/build_executable.py
@@ -95,12 +95,7 @@ if __name__ == "__main__":
             shutil.rmtree(repo_dir, ignore_errors=True)
 
         # clone repo with submodules
-        cmd = [
-            "git",
-            "clone",
-            "--recurse-submodules",
-            "https://github.com/isl-org/spear",
-            repo_dir]
+        cmd = ["git", "clone", "--recurse-submodules", "https://github.com/isl-org/spear", repo_dir]
         spear.log(f"Executing: {' '.join(cmd)}")
         subprocess.run(cmd, check=True)
 
@@ -108,11 +103,7 @@ if __name__ == "__main__":
         if args.commit_id:
             cwd = os.getcwd()
             os.chdir(repo_dir)
-            cmd = [
-                "git",
-                "reset",
-                "--hard",
-                args.commit_id]
+            cmd = ["git", "reset", "--hard", args.commit_id]
             spear.log(f"Executing: {' '.join(cmd)}")
             subprocess.run(cmd, check=True)
             os.chdir(cwd)
--- a/tools/build_executable.py
+++ b/tools/build_executable.py
@@ -95,7 +95,12 @@ if __name__ == "__main__":
             shutil.rmtree(repo_dir, ignore_errors=True)
 
         # clone repo with submodules
-        cmd = ["git", "clone", "--recurse-submodules", "https://github.com/isl-org/spear", repo_dir]
+        cmd = [
+            "git",
+            "clone",
+            "--recurse-submodules",
+            "https://github.com/isl-org/spear",
+            repo_dir]
         spear.log(f"Executing: {' '.join(cmd)}")
         subprocess.run(cmd, check=True)
 
@@ -103,7 +108,11 @@ if __name__ == "__main__":
         if args.commit_id:
             cwd = os.getcwd()
             os.chdir(repo_dir)
-            cmd = ["git", "reset", "--hard", args.commit_id]
+            cmd = [
+                "git",
+                "reset",
+                "--hard",
+                args.commit_id]
             spear.log(f"Executing: {' '.join(cmd)}")
             subprocess.run(cmd, check=True)
             os.chdir(cwd)
@@ -159,8 +168,7 @@ if __name__ == "__main__":
         "-target=SpearSim",
         "-archivedirectory=" + archive_dir,
         "-clientconfig=" + build_config,
-        run_uat_platform_args
-    ]
+        run_uat_platform_args]
     spear.log(f"Executing: {' '.join(cmd)}")
     subprocess.run(cmd, check=True)
 
--- a/tools/build_executable.py
+++ b/tools/build_executable.py
@@ -88,7 +88,7 @@ if __name__ == "__main__":
         shutil.rmtree(unreal_tmp_dir, ignore_errors=True)
 
     if not args.skip_clone_github_repo:
-        
+
         # remove our temporary repository directory to ensure a clean build
         if os.path.exists(repo_dir):
             spear.log(f"Repository exists, removing: {repo_dir}")
--- a/tools/build_paks.py
+++ b/tools/build_paks.py
@@ -16,7 +16,8 @@ import sys
 if __name__ == '__main__':
 
     parser = argparse.ArgumentParser()
-    parser.add_argument("--unreal_engine_dir", required=True)    
+    parser.add_argument("--unreal_engine_dir", required=True)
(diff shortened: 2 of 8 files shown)


tools/build_paks.py
```diff
@@ -16,7 +16,8 @@
 if __name__ == '__main__':
 
     parser = argparse.ArgumentParser()
-    parser.add_argument("--unreal_engine_dir", required=True)    
+    parser.add_argument("--unreal_engine_dir", required=True)
+    parser.add_argument("--unreal_project_dir", default=os.path.realpath(os.path.join(os.path.dirname(__file__), "..", "cpp", "unreal_projects", "SpearSim")))
```
mikeroberts3000, line 20: I just realized that our convention here is inconsistent. Some tools (e.g., `build_executable`) assume that `unreal_project_dir` is hard-coded relative to the `tools` folder. Other tools (e.g., `clean_project_dir`) have an argument for the project dir but set its default value.

I don't have a strong opinion about which convention we use, but let's use this PR as an opportunity to make it consistent across all the tools. Hard-coding is simpler and we only expect to have one project folder for the foreseeable future, but having an argument is probably more user-friendly because then users with their own projects can use our tools more easily. I think I like the convention you're using here, but let's apply it consistently to all the tools.

tools/upload_to_cloud.py
```diff
@@ -0,0 +1,45 @@
...
+import subprocess
+import sys
+
+
+if __name__ == "__main__":
+
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--cloud_path_prefix", required=True)
+    parser.add_argument("--version_tag", required=True)
+    parser.add_argument("--paks", action="store_true")
```
mikeroberts3000, line 14: The way `--paks` is used here won't scale if there are ever more than two different artifacts that we want to upload (i.e., executables, paks, and some other stuff). And there will be soon, because we will want to upload the results of our export pipeline. I recommend the following because then new artifacts can be added more easily without needing to change any surrounding logic.

```
    parser.add_argument("--upload_executable", action="store_true")
    parser.add_argument("--upload_paks", action="store_true")

    # ...

    if args.upload_executable:
        # ...
    if args.upload_paks:
        #
```

tools/upload_to_cloud.py
mikeroberts3000, line 1: The file is hardcoded to upload to AWS, so I think it's better if AWS is in the name of the file.

`upload_build_artifacts_to_aws.py`

mikeroberts3000 (approved):

tools/upload_to_cloud.py
RachithP: Aah, yes, this makes sense. I'll update this.

tools/build_paks.py
RachithP: I agree with this. I've made other files adhere to the convention in this file.
`build_executable` is a bit complicated as it is designed to build an executable after cloning from git repo, so as to assign a default project dir will incur more `if` statements and make the code a bit ugly. On the positive side, this script can be used to build executables from a working project dir, and this would be a bit more general purpose.

What do you think?

mikeroberts3000: Update: Actually this isn't as inconsistent as I thought. Currently, our policy seems to be, do it the hard-coded way unless there is a good reason not to. In the case of the scripts that are referred to in `build_executable`, we need to make them parameterized, because of the temp folder that is used in `build_executable`.

mikeroberts3000: Speaking of which, there is one thing that always bothered me with our approach for building pak files. There is some redundancy between the pak file that gets built by default with the executable, and the scene-specific pak files.

If it is easy to do so, we should remove all files from the scene-specific pak manifest that are already in the default manifest. Since running `build_executable` is a prerequisite for running `build_paks`, there should be the default pak manifest should be lying around somewhere in the project folder. Can we diff each scene-specific pak manifest against the default manifest so our pak files are smaller?

tools/upload_build_artifacts_to_aws.py
```diff
@@ -0,0 +1,52 @@
...
+        paks_filter_string = args.version_tag + "-" + "Mac.pak"
+    elif sys.platform == "linux":
+        platform_name = "Linux"
+        paks_filter_string = args.version_tag + "-" + "Linux.pak"
+    else:
+        assert False
+
+    aws_dir = posixpath.join(args.aws_path_prefix, args.version_tag, "")
+
+    files_to_upload = []
```
mikeroberts3000, line 39: It's better if the if statements are always listed in the same order.

```
    files_to_upload = []

    if args.upload_executable:
        files_to_upload.append(os.path.realpath(os.path.join(args.input_dir, f"SpearSim-{args.version_tag}-{platform_name}-Shipping.zip")))

    if args.upload_paks:
        files_to_upload.append(os.path.realpath(os.path.join(args.input_dir, x)) for x in os.listdir(args.input_dir) if x.endswith(paks_filter_string))
```

tools/build_executable.py
mikeroberts3000, line 1: We use this formatting convention in other tools and I like how it's more compact.

```
    cmd = [
        run_uat_script,
        "BuildCookRun",
        "-project=" + os.path.realpath(os.path.join(unreal_project_dir, "SpearSim.uproject")),
        "-build",            # build C++ executable
        "-cook",             # prepare cross-platform content for a specific platform (e.g., compile shaders)
        "-stage",            # copy C++ executable to a staging directory (required for -pak)
        "-package",          # prepare staged executable for distribution on a specific platform (e.g., runs otool and xcrun on macOS)
        "-archive",          # copy staged executable to -archivedirectory (on some platforms this will also move directories around relative to the executable)
        "-pak",              # generate a pak file for cooked content and configure executable so it can load pak files
        "-iterativecooking", # only cook content that needs to be updated
        "-targetplatform=" + target_platform,
        "-target=SpearSim",
        "-archivedirectory=" + archive_dir,
        "-clientconfig=" + build_config,
        run_uat_platform_args]
```

tools/build_executable.py
```diff
@@ -88,22 +88,31 @@
...
-        
+
         # remove our temporary repository directory to ensure a clean build
         if os.path.exists(repo_dir):
             spear.log(f"Repository exists, removing: {repo_dir}")
             shutil.rmtree(repo_dir, ignore_errors=True)
 
         # clone repo with submodules
-        cmd = ["git", "clone", "--recurse-submodules", "https://github.com/isl-org/spear", repo_dir]
+        cmd = [
```
mikeroberts3000, line 98: @RachithP I like it better the old way. If one of these `cmd` definitions can fit on one line, we should fit it on one line. If the `cmd` definition requires multiple lines, then I like the formatting in this file, but it shouldn't be mandatory that we put it on multiple lines if it's already so concise.

tools/prepare_scene_for_editing.py
```diff
@@ -11,10 +11,14 @@
 
     parser = argparse.ArgumentParser()
     parser.add_argument("--perforce_content_dir")
+    parser.add_argument("--unreal_project_dir", default=os.path.realpath(os.path.join(os.path.dirname(__file__), "..", "cpp", "unreal_projects", "SpearSim")))
```
mikeroberts3000, line 14: @RachithP what is the reason that `prepare_scene_for_editing` needs to be parameterized like this? We don't call it from `build_executable` or `build_paks`. Am I missing something?

tools/build_paks.py
RachithP: You are right, there are redundancies that we can avoid in building `pak` files. Let me check on this and get back.

PR #288, opened 2024-06-06
Make pak build process more efficient

RachithP:
* Reduce redundancy in per-scene pak files. This is achieved by removing common contents from per-scene pak file and packaging them into a `Common` pak file. This saves a lot of disk-space, and improves download and upload times.


tools/build_paks/build_paks.py
```diff
@@ -0,0 +1,186 @@
...
+
+    # create the output_dir
+    os.makedirs(output_dir, exist_ok=True)
+
+    # create manifest file in output dir
+    with open(txt_file, mode="w") as f:
+        for cooked_include_dir in cooked_include_dirs:
+            for cooked_include_file in glob.glob(os.path.realpath(os.path.join(cooked_include_dir, "**", "*.*")), recursive=True):
+                cooked_include_file_posix = cooked_include_file.replace(ntpath.sep, posixpath.sep)
+                mod_file =  [ cooked_include_file_posix[i:] for i in range(len(cooked_include_file_posix)) if cooked_include_file_posix.startswith("Engine" + posixpath.sep, i) ]
```
mikeroberts3000, line 75: This block of code is confusing. What is this code doing? Why do we need to explicitly materialize this array of all possible suffixes?

tools/build_paks/default_cooked_files.txt
mikeroberts3000, line 1: I don't think we should maintain this text file. This is going to be error-prone. Why can't we parse the manifest file that is generated in the `build_exectuable` step?

tools/build_paks/build_paks.py
mikeroberts3000:
```
    unreal_project_default_manifest_file = os.path.realpath(os.path.join(unreal_project_cooked_dir, "SpearSim", "Metadata", "packagestore.manifest"))

    # get asset files that are available by default
    with open(unreal_project_default_manifest_file) as f:
        default_manifest = json.load(f)
        default_manifest_asset_names = [ default_manifest_asset["Path"] for default_manifest_asset in default_manifest["Files"] ]

    # create manifest file in output dir
    with open(txt_file, mode="w") as f:
        for cooked_include_dir in cooked_include_dirs:
            for cooked_include_file in glob.glob(os.path.realpath(os.path.join(cooked_include_dir, "**", "*.*")), recursive=True):
                assert cooked_include_file.startswith(unreal_project_cooked_dir)
                asset_name = cooked_include_file.removeprefix(os.path.join(unreal_project_cooked_dir, os.path.sep)).replace(ntpath.sep, posixpath.sep)
                if asset_name not in default_manifest_asset_names:
                    cooked_include_file_posix = cooked_include_file.replace(ntpath.sep, posixpath.sep)
                    assert cooked_include_file_posix.startswith(unreal_project_cooked_dir_posix)
                    cooked_mount_file_posix = posixpath.join("..", "..", "..", cooked_include_file_posix.replace(unreal_project_cooked_dir_posix + posixpath.sep, ""))
                    f.write(f'"{cooked_include_file_posix}" "{cooked_mount_file_posix}" "" \n')
```

tools/build_paks.py
```diff
@@ -13,31 +14,102 @@
...
+        "-stdout",                               # ensure log output is written to the terminal 
+        "-fullstdoutlogoutput",                  # ensure log output is written to the terminal
+        "-nologtimes"]                           # don't print timestamps next to log messages twice
+    spear.log(f"    Executing: {' '.join(cmd)}")
+    subprocess.run(cmd, check=True)
+
+    # create the output_dir
+    os.makedirs(output_dir, exist_ok=True)
+
+    # create manifest file in output dir
```
mikeroberts3000, line 71: This code is getting a bit too dense so I just added some whitespace.

```
    # create manifest file in output dir
    with open(txt_file, mode="w") as f:
        for cooked_include_dir in cooked_include_dirs:
            for cooked_include_file in glob.glob(os.path.realpath(os.path.join(cooked_include_dir, "**", "*.*")), recursive=True):

                cooked_include_file_posix = cooked_include_file.replace(ntpath.sep, posixpath.sep)
                asset_name = cooked_include_file_posix.removeprefix(unreal_project_cooked_dir_posix + posixpath.sep)

                if asset_name not in default_manifest_asset_names:
                    assert cooked_include_file_posix.startswith(unreal_project_cooked_dir_posix)
                    cooked_mount_file_posix = posixpath.join("..", "..", "..", cooked_include_file_posix.replace(unreal_project_cooked_dir_posix + posixpath.sep, ""))
                    f.write(f'"{cooked_include_file_posix}" "{cooked_mount_file_posix}" "" \n')
```

tools/build_paks.py
```diff
@@ -53,135 +125,58 @@
...
-        # to the directory inside the user's Perforce workspace. Instead, we want this path to refer to the symlinked
-        # version inside the user's Unreal project directory.
-        perforce_dirs_to_unreal_dirs = {
+    unreal_project_content_dir           = os.path.realpath(os.path.join(unreal_project_dir, "Content"))
+    unreal_project_cooked_dir            = os.path.realpath(os.path.join(unreal_project_dir, "Saved", "Cooked", platform))
+    unreal_project_content_scenes_dir    = os.path.realpath(os.path.join(unreal_project_content_dir, "Scenes"))
+    unreal_project_default_manifest_file = os.path.realpath(os.path.join(unreal_project_cooked_dir, "SpearSim", "Metadata", "packagestore.manifest"))
+    perforce_content_scenes_dir          = os.path.realpath(os.path.join(args.perforce_content_dir, "Scenes"))
+
+    # get asset names that are available by default
```
mikeroberts3000, line 134: I just renamed the inner variable in the list comprehension to make things a bit more concise.

```
    # get asset names that are available by default
    assert os.path.exists(unreal_project_default_manifest_file)
    with open(unreal_project_default_manifest_file) as f:
        default_manifest = json.load(f)
        default_manifest_asset_names = [ asset["Path"] for asset in default_manifest["Files"] ]
```

tools/build_paks.py
```diff
@@ -53,135 +125,58 @@
...
-            scene_ids = [ s for s in scene_ids if fnmatch.fnmatch(s, args.scene_ids) ]
-        assert len(scene_ids) > 0
-
-    # Create a symlink to common top-level directories
-    if not args.skip_create_symlinks:        
-        # We do not want to use os.path.realpath(...) for the values in this dictionary, because that will resolve
-        # to the directory inside the user's Perforce workspace. Instead, we want this path to refer to the symlinked
-        # version inside the user's Unreal project directory.
-        perforce_dirs_to_unreal_dirs = {
+    unreal_project_content_dir           = os.path.realpath(os.path.join(unreal_project_dir, "Content"))
```
mikeroberts3000, line 128:
```
    if sys.platform == "win32":
        platform          = "Windows"
        unreal_editor_bin = os.path.realpath(os.path.join(args.unreal_engine_dir, "Engine", "Binaries", "Win64", "UnrealEditor.exe"))
        unreal_pak_bin    = os.path.realpath(os.path.join(args.unreal_engine_dir, "Engine", "Binaries", "Win64", "UnrealPak.exe"))
    elif sys.platform == "darwin":
        platform          = "Mac"
        unreal_editor_bin = os.path.realpath(os.path.join(args.unreal_engine_dir, "Engine", "Binaries", "Mac", "UnrealEditor.app", "Contents", "MacOS", "UnrealEditor"))
        unreal_pak_bin    = os.path.realpath(os.path.join(args.unreal_engine_dir, "Engine", "Binaries", "Mac", "UnrealPak"))
    elif sys.platform == "linux":
        platform          = "Linux"
        unreal_editor_bin = os.path.realpath(os.path.join(args.unreal_engine_dir, "Engine", "Binaries", "Linux", "UnrealEditor"))
        unreal_pak_bin    = os.path.realpath(os.path.join(args.unreal_engine_dir, "Engine", "Binaries", "Linux", "UnrealPak"))
    else:
        assert False

    unreal_project_dir                   = os.path.realpath(args.unreal_project_dir)
    unreal_project_content_dir           = os.path.realpath(os.path.join(unreal_project_dir, "Content"))
    unreal_project_content_scenes_dir    = os.path.realpath(os.path.join(unreal_project_content_dir, "Scenes"))
    unreal_project_cooked_dir            = os.path.realpath(os.path.join(unreal_project_dir, "Saved", "Cooked", platform))
    unreal_project_default_manifest_file = os.path.realpath(os.path.join(unreal_project_cooked_dir, "SpearSim", "Metadata", "packagestore.manifest"))
    perforce_content_scenes_dir          = os.path.realpath(os.path.join(args.perforce_content_dir, "Scenes"))

    # if you want to assert paths exist, do it here, but I don't think it's necessary
```

mikeroberts3000 (approved):
Please address my latest comments but otherwise this looks great.

tools/build_paks/default_cooked_files.txt
RachithP: There is one issue in this method - since the `packagestore.manifest` files gets updated every time we cook (new entries from new scenes gets added), we won't be able to restart the `build_paks.py` procedure for an old scene as the `packagestore.manifest` file would already contain entries for that scene and our current code will exclude these files during `pak` creation leading to incomplete `pak` file.

tools/build_paks/build_paks.py
RachithP: Aah, forgot about the `removeprefix` functionality. It's good that we also recently upgraded to python 3.9 as `removeprefix` was introducted in pythobn 3.9.

tools/build_paks/default_cooked_files.txt
RachithP: In order to navigate this, I create a backup file of `packagestore.manifest` the first time we run this script, and use that for every subsequent runs.

tools/build_paks/build_paks.py
mikeroberts3000: In Python 3.8 you could also do `assert str.startswith(bad_str); str.replace(bad_str, "", 1)` to indicate that you only want to replace remove the 1st occurance of `bad_str`. But I agree that `removeprefix` is cleaner.

tools/build_paks/default_cooked_files.txt
mikeroberts3000: Ah, that is an important detail. I think it's brittle to make a copy of `packagestore.manifest` the first time we run this script, because how will the script know it is being run for the first time? I think it's better to make a copy of `packagestore.manifest` at the end of `build_executable` (e.g., in the same folder but named `packagestore.manifest.default`).

Then we don't need any weird cleanup logic when things are run repeatedly. `build_executable` always generates `packagestore.manifest.default`, and `build_paks` always reads `packagestore.manifest.default`.

mikeroberts3000: I like your latest solution of using the `UnrealPak -List` option to get the default manifest in a stable way. Of all the options we've considered so far, this is the simplest.

tools/build_paks.py
```diff
@@ -13,30 +13,104 @@
...
     parser = argparse.ArgumentParser()
     parser.add_argument("--unreal_engine_dir", required=True)
-    parser.add_argument("--unreal_project_dir", default=os.path.realpath(os.path.join(os.path.dirname(__file__), "..", "cpp", "unreal_projects", "SpearSim")))
     parser.add_argument("--output_dir", required=True)
     parser.add_argument("--version_tag", required=True)
-    parser.add_argument("--skip_create_symlinks", action="store_true")
-    parser.add_argument("--perforce_content_dir")
+    parser.add_argument("--perforce_content_dir", required=True)
+    parser.add_argument("--unreal_project_dir", default=os.path.realpath(os.path.join(os.path.dirname(__file__), "..", "cpp", "unreal_projects", "SpearSim")))
+    parser.add_argument("--spearsim_executable_dir", default=os.path.realpath(os.path.join(os.path.dirname(__file__), "build", "SpearSim-Win64-Shipping")))
```
mikeroberts3000, line 105: Let's call this `build_dir` because that is what it's called in the `build_executable` tool, and then let's set the full path in our platform-specific logic.

tools/build_paks/default_cooked_files.txt
RachithP: We don't need to have to worry about this issue now because there is a much simpler way - using the default `pak` file (SpearSim-<platform>.pak`) that gets shipped with the executable. The `packagestore.manifest` file appears to be a subset of files in `SpearSim-<platform>.pak`, so we need to use `SpearSim-<package>.pak` to maximize our reduction in redundancy. As an added bonus, the code in this file is simpler and cleaner.

tools/build_paks.py
RachithP: Aah, right. Missed the `Win64` in the path.

tools/build_paks/default_cooked_files.txt
mikeroberts3000: 100% agreed with your reasoning here, and I like how you clearly articulated it in your post above. I might even go so far as to say the simplified code is the main selling point, and the reduced redundancy is the added bonus. I also appreciate that you deviated from my latest suggestion and came up with an alternative approach that is better in every way (cleaner, simpler, less redundancy).

I wasn't aware that the `packagestore.manifest` manifest file only contains a subset of the stuff in `SpearSim-platform.pak`. I thought you did that experiment and decided that the contents of each were equivalent. Did you discover some extra files that are in `SpearSim-platform.pak` but not in `packagestore.manifest`? Why did we originally think the contents of each were equivalent but now we no longer think so?

RachithP: You are right, there was an error in my experiments. In my initial experiments, I compared only the `.uasset` files, and concluded that the files in both `packagestore.manifest` and `SpearSim-<platform>.pak` were same. I had skipped `.ubulk` and `.uexp` files during my filtering step. Recently I discovered this error, and found that  the `.ubulk` and `.uexp` files account for a significant portion of the reduced file size.

PR #290, opened 2024-06-13
add `download_paks.py` script and dependencies

RachithP:


tools/download_paks.py
```diff
@@ -0,0 +1,63 @@
...
+from tqdm import tqdm
+import urllib.request
+
+if __name__ == '__main__':
+
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--output_dir", required=True)
+    parser.add_argument("--platform", required=True)
+    parser.add_argument("--version_tag", required=True)
+    parser.add_argument("--scene_ids", required=True)
```
mikeroberts3000, line 19: - `scene_ids` should be optional, if it is not given then the script should download everything
- it is ok if all possible scene names are hard-coded in this script somewhere

tools/download_paks.py
```diff
@@ -0,0 +1,63 @@
...
+    args = parser.parse_args()
+
+    assert args.platform in ["Windows", "Linux", "Mac"]
+
+    # create output dir
+    output_dir = os.path.realpath(args.output_dir)
+    os.makedirs(output_dir, exist_ok=True)
+
+    # create a list of available scene_ids
+    available_scene_ids = [ f"kujiale_{str(i).zfill(4)}" for i in range(30) ]
```
mikeroberts3000, line 29: It would be better to materialize this list explicitly because it is ugly that we handle `warehouse_0000` differently from `kujiale_*`, and we want to be able to skip Kujiale scenes without breaking the logic here.

tools/download_paks.py
```diff
@@ -0,0 +1,63 @@
...
+    for scene_id in scene_ids:
+
+        pak_file_name = scene_id + "-" + args.version_tag + "-" + args.platform + ".pak"
+        download_url  = posixpath.join(download_url_prefix, args.version_tag, pak_file_name)
+        output_file   = os.path.join(output_dir, pak_file_name)
+
+        spear.log(f"Downloading {download_url} to {output_file}")
+
+        # create a progress bar
+        response = urllib.request.urlopen(download_url)
```
mikeroberts3000, line 49: This downloading code seems really low-level. Is this really the cleanest simplest way to download a file? See the Hypersim download script for an example how to do this with a single line of code.

RachithP: There is a simpler way to do it using `urllib.request.urlretrieve(url, file)`. But, this doesn't show the download progress. I added this extra bit of logic to show the download progress.

mikeroberts3000: Fair enough. Let's remove the white space to indicate that it's all one big block that the reader shouldn't pay too much attention to. And let's use " instead of ' because that is our convention unless we have a good reason not to (if the string contains the " character, that is a good reason to use ' to enclose).

RachithP: Actually, there's a simpler way to track progress using `wget`. I updated the code to reflect this. Instead of `tqdm` we add `wget` dependency and the code is much simpler now.

mikeroberts3000: Sweet. Can we use `curl` on Windows as a fallback? I'm not sure if `wget` is available on Windows.

mikeroberts3000: Update: Yes, the `wget` Python package is available on Windows.

tools/download_paks.py
```diff
@@ -45,19 +43,6 @@
...
-        with open(output_file, 'wb') as file:
-            while True:
-                buffer = response.read(block_size)
-                if not buffer:
-                    break
-
-                file.write(buffer)
-                tqdm_bar.update(len(buffer))
-        tqdm_bar.close()
+        wget.download(download_url, out=output_file)
```
mikeroberts3000, line 82: Much better 😎

mikeroberts3000 (approved):

PR #292, opened 2024-06-14
update `sign_executable.py`

RachithP:


tools/sign_executable/sign_executable.py
```diff
@@ -87,6 +87,7 @@
         "ditto",
         "-c",
         "-k",
+        "--sequesterRsrc",
```
mikeroberts3000, line 90: There is a comment above this line of code that explains what all the arguments do. Please update the comment to explain what this new arg is doing.

mikeroberts3000 (approved):

PR #293, opened 2024-06-15
update to reflect v0.5.0 scenes

RachithP:


tools/download_paks.py
```diff
@@ -12,17 +12,16 @@
 
 # list all available scene_ids for download
 available_scene_ids = [
+    "Common",
```
mikeroberts3000, line 15: Can we rename this to `common` so it matches all the other pak files? The filename of each pak file doesn't matter to Unreal, so we wouldn't need to rebuild anything. We would just need to rename the pak files on AWS, and change the default name in `build_paks` so it produces the correct output the next time we need to run it.

tools/download_paks.py
```diff
@@ -12,17 +12,16 @@
...
     "kujiale_0000",
     "kujiale_0001",
     "kujiale_0002",
     "kujiale_0003",
     "kujiale_0004",
-    "kujiale_0005",
```
mikeroberts3000, line 20: Let's do comments rather than removing entirely. It will hint to the reader that the missing scenes will be back soon.

mikeroberts3000 (approved):

PR #294, opened 2024-06-17
use Unreal shipped clang and libc++ on Linux

RachithP:


mikeroberts3000 (approved):

mikeroberts3000:
Awesome! I simplified the build instructions to match the changes in this PR 😎

mikeroberts3000 (approved):

tools/build_third_party_libs.py
```diff
@@ -53,7 +68,7 @@
         cxx_flags = "'-std=c++20 -mmacosx-version-min=10.14'"
     elif sys.platform == "linux":
         platform_dir = "Linux"
-        cxx_flags = "'-std=c++20 -stdlib=libc++'"
+        cxx_flags = f"'-std=c++20 -stdlib=libc++ -nostdinc++ -I{linux_libcpp_include_dir} -L{linux_libcpp_lib_dir} -Qunused-arguments'"
```
mikeroberts3000, line 71: - I understand why `-nostdinc++` would be necessary here. But is it necessary to specify `-stdlib=libc++` when specifying `-nostdinc++`? It seems weird that we would need to specify both.
- I understand that `cxx_compiler` is invoked by this script, but is `c_compiler` ever invoked? If not, we should get rid of it. I'm thinking it might not be, because we never specify `c_flags`.
- Let's not try to suppress warnings in third-party code.
  - It seems like a lot of work to have this policy for all third-party libs, and if it is not worth doing consistently, it is not worth doing at all.
  - I don't mind seeing the warnings, especially when building third-party libs, which is a one-time operation.
  - In general, we don't understand third-party code well enough to decide if we should really be suppressing warnings or not. I know we do this sometimes in our code, e.g., when including third-party headers. But that is because we would see those warnings when building our code, so there is a risk of too many third-party warnings (which we might not care about) hiding warnings generated by our code (which we should definitely care about). In this case, we are not building any of our code, so there is no such danger.
- We should only add these extra flags if the user didn't specify `args.cxx_compiler`.

mikeroberts3000: - Based on my experiments using `-DCMAKE_VERBOSE_MAKEFILE=ON`, I have concluded that `c_compiler` is never invoked by this script. 100% of compiled files are compiled with `cxx_compiler`. We should therefore remove `c_compiler` from this script entirely because it will simplify some of the downstream logic.
- Let's include a `--verbose` option that adds `-DCMAKE_VERBOSE_MAKEFILE=ON` to every `cmake` invocation (excluding the `cmake -build` invocations).
- Let's sort the `cmake` arguments consistently as follows: `DCMAKE_CXX_COMPILER`, `DCMAKE_CXX_FLAGS`, `DCMAKE_PLATFORM_SPECIFIC_STUFF`, `DCMAKE_VERBOSE_MAKEFILE`.

RachithP: 1. `-nostdinc++` is required as we don't want to use the standard library headers available in the default paths of the system. So, `-nostdinc++` forces cmake to search for the headers in only directories specified by `-I{}`. However, `-stdlib=libc++` is required as the default standard library on linux is `libstdc++` and clang uses it unless `libc++` is specified.
2. I'll perform a test by removing `c_compiler`, and remove it if not needed.
3. It's weird that adding these extra flags to yaml-cpp build throws `argument unused during compilation` warning for `stdlib=libc++`, `-L{linux_libcpp_lib_dir}`, and `-nostdinc++`, and when these flags are removed (and only `-std=c++20 -I{linux_libcpp_include_dir}` is kept), there are linker errors.

tools/build_third_party_libs.py
```diff
@@ -14,36 +14,30 @@
 
     parser = argparse.ArgumentParser()
     parser.add_argument("--third_party_dir", default=os.path.realpath(os.path.join(os.path.dirname(__file__), "..", "third_party")))
+    parser.add_argument("--unreal_engine_dir")
     parser.add_argument("--num_parallel_jobs", type=int, default=1)
-    parser.add_argument("--c_compiler")
     parser.add_argument("--cxx_compiler")
+    parser.add_argument("--verbose", action="store_true")
     args = parser.parse_args()
```
mikeroberts3000, line 21:
```
    parser.add_argument("--third_party_dir", default=os.path.realpath(os.path.join(os.path.dirname(__file__), "..", "third_party")))
    parser.add_argument("--unreal_engine_dir")
    parser.add_argument("--num_parallel_jobs", type=int, default=1)
    parser.add_argument("--cxx_compiler")
    parser.add_argument("--verbose", action="store_true")
    args = parser.parse_args()

    assert os.path.exists(args.third_party_dir)
    third_party_dir = os.path.realpath(args.third_party_dir)
```

tools/build_third_party_libs.py
```diff
@@ -209,13 +215,13 @@
         cmd = ["cmake", "--build", ".", "--config", "Release", "-j", f"{args.num_parallel_jobs}"]
 
     elif sys.platform == "linux":
-
         cmd = [
```
mikeroberts3000, line 218: Grouping `CMAKE` vars and `YAML_CPP` vars:

```
        cmd = [
            "cmake",
            "-DCMAKE_CXX_COMPILER=" + cxx_compiler,
            "-DCMAKE_CXX_FLAGS=" + cxx_flags,
            "-DCMAKE_POSITION_INDEPENDENT_CODE=ON",
            "-DCMAKE_VERBOSE_MAKEFILE=" + verbose_makefile,
            "-DYAML_CPP_BUILD_TESTS=OFF",
            os.path.join("..", "..")]
```

tools/build_third_party_libs.py
mikeroberts3000: Cool.

tools/build_third_party_libs.py
```diff
@@ -53,10 +47,23 @@
         cxx_flags = "'-std=c++20 -mmacosx-version-min=10.14'"
     elif sys.platform == "linux":
         platform_dir = "Linux"
-        cxx_flags = "'-std=c++20 -stdlib=libc++'"
+        cxx_flags = "-std=c++20 -stdlib=libc++"
```
mikeroberts3000, line 50: Are the quotation rules different on Linux? That seems weird. Also, if you don't want to deal with appending to `cxx_flags`, it wouldn't be the end of the world to just re-assign to it.

```
if whatever:
    cxx_flags = "-std=c++20 -stdlib=libc++"

...

if something_else:
    cxx_flags = "-std=c++20 -stdlib=libc++ -bunch -of -extra -stuff"
```

tools/build_third_party_libs.py
RachithP: I like (a) too. I've updated code to make it consistent across platforms.

PR #304, opened 2024-12-06
Adds camera buffers

soravux:
Adds the following native buffers from Unreal as possible output when rendering:

- Diffuse Color
- Specular Color
- Roughness
- Metallic
- Ambient Occlusion


pipeline/create_blueprint_bp_camera_sensor.py
```diff
@@ -50,6 +50,61 @@
             "material_path": "/SpComponents/Materials/PPM_Normal.PPM_Normal",
             "fov_angle": 90.0,
         },
+        {
```
mikeroberts3000, line 53: - Sort modalities alphabetically by name.

mikeroberts3000 (approved):
Please address my comments, either by commenting or modifying the PR or both, and feel free to check in when ready 🥳

pipeline/create_blueprint_bp_camera_sensor.py
soravux: Thanks for the feedback!

- Sort modalities: Done in the next commit.
- PPM material complexity: I checked them all with "show only modified properties"; they have the same or lower complexity as the existing nodes.
- Raw outputs from Unreal: No buffer in this PR results from a transformation; they are all direct Unreal outputs from the postprocess pass mapped to the emissive surface output of the shader.
- Commit `BP_Camera_Sensor`: Done in the next commit.
- Name concordance between materials and textures: All are named the same, modulo SnakeCase and underscores.
- I tested them by adapting the `render_image` example, and it gives the following results, which look good to me:

unreal_buffers_test

PR #319, opened 2026-07-10
Add a minimal Path Tracing example

jakubtomsu:
This PR implements a simple SPEAR program to show how to use Unreal's built-in offline Path Tracer. This is done purely with Scene Capture Component, without using Movie Render Queue to maximize flexibility and control.

The code is quite simple; all it needs is a few *cvar* overrides and a *showflag* to let the Scene Capture Component use PT.

At runtime it behaves almost exactly like the in-editor PT viewport - samples are accumulated into a framebuffer until a maximum sample count is reached. If the camera or something in the scene moves all previous samples are invalidated and the render starts over. For this reason there is a `--manual-accumulate` flag that reads pixels and flushes after every sample, but it's somewhat slower.

~~There is also support for HDR if you specify an `.exr` output path.~~


examples/render_image_path_tracing/run.py
```diff
@@ -0,0 +1,170 @@
...
+#   - Ray tracing is enabled for the scene-capture view. FSceneView::IsRayTracingAllowedForView() allows ray
+#     tracing for a scene capture only when bSceneCaptureUsesRayTracing is set, which comes from
+#     r.RayTracing.SceneCaptures or the component's bUseRayTracingIfEnabled.
+#   - The PathTracing engine show flag is set on the view (selects the path tracer over the deferred renderer).
+
+
+import argparse
+import os
+
+os.environ.setdefault("OPENCV_IO_ENABLE_OPENEXR", "1")  # enable cv2's EXR codec; must be set before importing cv2
```
mikeroberts3000: To keep this example simple, let's not bother saving HDR. So we can remove this.

examples/render_image_path_tracing/run.py
```diff
@@ -0,0 +1,170 @@
...
+import cv2
+import spear
+import numpy as np
+
+_NUM_EXTRA_SETTLE_FRAMES = 4
+
+if __name__ == "__main__":
+
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--out", type=str, default=os.path.realpath(os.path.join(os.path.dirname(__file__), "out.png")))
```
mikeroberts3000: For simplicity and consistency with other examples, you can just hardcode the filename to `image.png`, and use the same approach as in `render_image` to set width and height. I think samples, bounces, and denoiser should be specified in `user_config.yaml` in a `CONFIG_ENGINE_INI_STRING` block.

examples/render_image_path_tracing/run.py
```diff
@@ -0,0 +1,170 @@
...
+    parser.add_argument("--width", type=int, default=None) # Viewport by default
+    parser.add_argument("--height", type=int, default=None) # Viewport by default
+    parser.add_argument("--samples", type=int, default=64)
+    parser.add_argument("--bounces", type=int, default=8) # More than ~12 bounces makes almost no difference in most scenes.
+    parser.add_argument("--denoiser", type=str, default=None)  # e.g. "NNEDenoiser" or "OIDN"; plugin must be enabled. No denoising by default.
```
mikeroberts3000: Do these plugins ship with the engine? If so, then let's identify all reasonable denoiser options that would be useful in this example, all the plugins (assuming they ship with the engine) that would be required to use those denoisers, and enable all of those plugins in `SpearSim.uproject`.

examples/render_image_path_tracing/run.py
```diff
@@ -0,0 +1,170 @@
...
+        exit(1)
+
+    # create instance
+    config = spear.get_config(user_config_files=[os.path.realpath(os.path.join(os.path.dirname(__file__), "user_config.yaml"))])
+
+    spear.configure_system(config=config)
+    instance = spear.Instance(config=config)
+    game = instance.get_game()
+
+    is_hdr = str.endswith(args.out, ".exr")
```
mikeroberts3000: Remove.

examples/render_image_path_tracing/run.py
```diff
@@ -0,0 +1,170 @@
...
+
+    spear.log("Info: Starting ...")
```
mikeroberts3000: Move this to when initialization and configuration is complete and the rendering is actually beginning.

```
spear.log("Path-traced rendering beginning...")
```

examples/render_image_path_tracing/run.py
```diff
@@ -0,0 +1,170 @@
...
+
+
+    # Initialize
+
+    with instance.begin_frame():
+
+        scene_captures_cvar = game.unreal_service.find_console_variable_by_name(console_variable_name="r.RayTracing.SceneCaptures")
```
mikeroberts3000: I think all of these CVars should be set in `user_config.yaml` under a `CONFIG_ENGINE_INI_STRING` block.

examples/render_image_path_tracing/run.py
```diff
@@ -0,0 +1,170 @@
...
+        game.rendering_service.align_camera_with_viewport(camera_sensor=bp_camera_sensor, camera_components=scene_capture_component, viewport_desc=viewport_desc, widths=args.width, heights=args.height)
+
+        # enable ray tracing for this component and turn on the PathTracing show flag.
+        # SetShowFlagSettings() applies immediately.
+        scene_capture_component.bUseRayTracingIfEnabled = True
+        scene_capture_component.SetShowFlagSettings(InShowFlagSettings=[
+            {"ShowFlagName": "PathTracing", "Enabled": True},
+            {"ShowFlagName": "PostProcessing", "Enabled": False},
+        ])
+        scene_capture_component.CaptureSource = "SCS_FinalColorHDR" # no tonemapping
```
mikeroberts3000: Disabling all tonemapping requires more care than this in the Lumen code path, so I'm not sure this code is correct. But I also don't think we need to disable tonemapping for this example.

Can we keep tone-mapping enabled so the example behaves just like `render_image` but the only thing we're changing is Lumen vs PT? See `create_asset_bp_camera_sensor.py` for details about what we need to disable tonemapping, e.g., in the Lumen code path, the only way I've found to completely disable all tonemapping effects is to access the `PPM_PostProcessInput2` post-processing material.

examples/render_image_path_tracing/run.py
```diff
@@ -0,0 +1,170 @@
...
+    # Render image
+
+    if args.manual_accumulate:
+        # manual accumulation: average independent 1-spp frames in Python, forcing a reset each frame.
+        # Works even when the scene invalidates the engine's own accumulation every frame (e.g. animated content)
+        accumulator = None
+        for i in range(args.samples):
+            spear.log(f"Info: Path tracing sample {i + 1}/{args.samples} ...")
+            with instance.begin_frame():
+                kismet_rendering_library = game.get_unreal_object(uclass="UKismetRenderingLibrary")
```
mikeroberts3000: `get_unreal_object` is a relatively heavy operation (i.e., it results in multiple synchronous RPC calls) and therefore should only be done once in your initialization code.

examples/render_image_path_tracing/run.py
```diff
@@ -0,0 +1,170 @@
...
+                kismet_rendering_library.RefreshPathTracingOutput() # reset accumulation before this frame renders
```
mikeroberts3000: I don't understand what `RefreshPathTracingOutput()` does, why it is necessary here, and under what circumstances the engine will do it's own refresh. The comment is confusing. It's fine if this example doesn't support manual accumulation, because manual accumulation would require correctly disabling tonemapping and I'm not sure we have done that correctly here.

I will only be convinced we are correctly disabling all tonemapping if we implement the following test:

```
image = ...
albedo = ...
irradiance = ...
residual = ...

assert np.allclose(image, albedo*irradiance + residual)
```

We have done this for carefully for Lumen code path, but I'm not sure how to obtain `irradiance` and `residual` for the PT code path.

examples/render_image_path_tracing/run.py
```diff
@@ -0,0 +1,170 @@
...
+            with instance.end_frame():
+                sample = scene_capture_component.read_pixels()["arrays"]["data"].astype(np.float32)
+            accumulator = sample if accumulator is None else accumulator + sample
+        final_pixels = accumulator / args.samples
+
+    else:
+        # Accumulate one path-traced sample per pixel per frame; r.PathTracing.SamplesPerPixel caps this at args.samples.
+        for i in range(0, args.samples):
```
mikeroberts3000: - `for i in range(args.samples)` for consistency. No need to specify 0 as the starting point.

- I think `args.samples` should be called `args.num_frames` or something because "sample" is a confusing overloaded term here that could mean "intermediate image" and also "result from a single primary ray". Confusing.

- Also, when exactly does denoising get applied? The comment below makes it sound like the denoiser only applies after we've called `step()` some number of times. But how would the engine possibly know to apply the denoiser after this for loop is finished executing? If the engine is accumulating PT data across frames and then applying denoising after some number of frames, the comments in this example should make it more clear how this works.

examples/render_image_path_tracing/run.py
```diff
@@ -0,0 +1,170 @@
...
+        # get rendered frame
+        with instance.begin_frame():
+            pass
+        with instance.end_frame():
+            final_pixels = scene_capture_component.read_pixels()["arrays"]["data"]
+            
+
+    # Save render
+
+    spear.log("Info: Saving image: ", args.out)
```
mikeroberts3000: This image saving code could be simplified if we drop support for manual accumulation.

examples/render_image_path_tracing/run.py
```diff
@@ -0,0 +1,170 @@
...
+        # Wait for the final pass and possibly the denoiser
+        instance.step(num_frames=_NUM_EXTRA_SETTLE_FRAMES)
+
+        # get rendered frame
+        with instance.begin_frame():
+            pass
+        with instance.end_frame():
+            final_pixels = scene_capture_component.read_pixels()["arrays"]["data"]
+            
+
```
mikeroberts3000: Symmetric comment with the comment above:

```
spear.log("Path-traced rendering finished.")
```

examples/render_image_path_tracing/run.py
jakubtomsu: Okay, the tonemapping makes sense, I'll just keep the default tonemapped output. Postprocessing can remain too for sure, though in the case of pathtracing some postprocess effects can amplify noise by a lot if you aren't using a denoiser. For example, chromatic aberration almost sharpens the noise and on low SPP creates these weird grid-looking color artifacts (due to the way the chromatic aberration strength is computed per pixel probably)

examples/render_image_path_tracing/run.py
jakubtomsu: That doesn't work in the case of `LAUNCH_MODE: "none"` though, correct? Or am I missing something? What we need is to actually override the flags at runtime.

jakubtomsu:
Okay I updated the code based on the comments above. I didn't see a good way to implement everything though, for instance many of the CVars are set at runtime because that works with `LAUNCH_MODE:"none"`. In this mode the ini string is ignored.
Another reason why I strongly prefer the commandline parameters for things like `num-frames` as opposed to storing them in `CONFIG_ENGINE_INI_STRING` is to make sure they stay in sync (it doesn't make sense to have a different value in config yaml and the commandline value. It would either render not enough frames and not denoise, or render more frames for no reason)

Regarding the extra step (the settle frames), I remember running into issues where I stepped the correct number of frames but the result wasn't denoised for some reason, and a small delay fixed it. Not sure what it was caused by and now it doesn't seem to be happening, so it was probably an unrelated issue, and the settle steps can be safely removed.

I got rid of the postprocessing override and the EXR output, however I think it makes sense to still manually disable the camera imperfections postprocess pass, since it produces bad artifacts (the chromatic aberration to be specific). That was the original intent with disabling the postprocessing. Default tonemapping is also back.

The manual accumulation is now also removed. I tested before it and it looked okay, but you're right something like this probably needs to be tested more rigorously to make sure the output is 100% correct before releasing it as an official example.

cpp/unreal_projects/SpearSim/SpearSim.uproject
mikeroberts3000, line 1: This is good but don't change the whitespace for no reason. Just make the minimal change that implements what you want.

cpp/unreal_projects/SpearSim/SpearSim.uproject
```diff
@@ -48,10 +45,17 @@
         {
             "Name": "SpCore",
             "Enabled":  true
+        },
+        {
```
mikeroberts3000, line 48: Place plugins in alphabetical order.

examples/render_image_path_tracing/README.md
```diff
@@ -0,0 +1,23 @@
+# Render a Path-Traced Image
```
mikeroberts3000, line 1: Since we're adding a new example, add an entry to `docs/running_our_example_applications.md`.

Make this documentation mimic the structure and overall voice of our existing example `README` files as closely as possible. It reads like it was written by a different person than all of the other examples. For example, we don't usually mention in these READMEs that you need to rename the `.example` file, because we already mention it once in a central location in our _Running our Example Applications_ doc, so mentioning it in every `README` is redundant and would make the `README` files harder to maintain.

I'm fine with `--teaser`, `--num-frames`, `--bounces`, and `--denoiser` being command-line arguments, but there are some cvars that can't be changed at runtime (e.g., `r.RayTracing.Enable=1`). For these cvars, instead of requiring the user to manually change `DefaultEngine.ini`, let's put all cvars that don't need to change based on the command-line arguments in an INI string in `user_config.yaml`.

I think it is even more powerful and non-obvious for this example to show that path-tracing can be enabled in a standalone executable. So I think `LAUNCH_MODE == "game"` should be the default. But in case a user wants to run the example in the editor with `LAUNCH_MODE == "none"`, it is easy to launch the editor using our `run_editor.py` tool, which accepts an optional `user_config.yaml` file, and avoids the need to manually edit `DefaultEngine.ini`. This is how our other examples work when we need to launch the editor with some custom INI values.

examples/render_image_path_tracing/README.md
```diff
@@ -0,0 +1,23 @@
...
+- `--teaser` — render at 1920x1080 (otherwise the viewport resolution is used).
+- `--num-frames` — number of frames to render while the path tracer accumulates samples (more frames, less noise).
+- `--bounces` — maximum path-tracer bounces.
+- `--denoiser` — spatial denoiser to apply, e.g. `NNEDenoiser` or `OIDN` (off by default).
+
+## Before running
+
+- A hardware ray-tracing capable GPU is required. Set `r.RayTracing.Enable=1` in SpearSim's `DefaultEngine.ini` (it must be set at startup and can't be changed at runtime).
+- Rename `user_config.yaml.example` -> `user_config.yaml`.
+- The example config uses `LAUNCH_MODE:"none"` + `EDITOR_LAUNCH_MODE:"game"`: open the SpearSim project in the editor, enter play mode, then run `run.py`. The render uses the viewport camera and the current level.
```
mikeroberts3000: `EDITOR_LAUNCH_MODE` only has an effect if `LAUNCH_MODE == "editor"`.

examples/render_image_path_tracing/run.py
```diff
@@ -0,0 +1,131 @@
...
+import argparse
+import cv2
+import os
+import spear
+
+parser = argparse.ArgumentParser()
+parser.add_argument("--teaser", action="store_true")
+parser.add_argument("--num-frames", type=int, default=64)
+parser.add_argument("--bounces", type=int, default=8) # More than ~12 bounces makes almost no difference in most scenes.
+parser.add_argument("--denoiser", type=str, default=None) # e.g. "NNEDenoiser" or "OIDN"; the denoiser's plugin must be enabled in SpearSim.uproject. No denoising by default.
```
mikeroberts3000: Comment not up-to-date.

examples/render_image_path_tracing/run.py
mikeroberts3000: I think `LAUNCH_MODE == "game"` should be the default. If a user wants to use `LAUNCH_MODE == "none"`, we should require the user to launch the editor via our `run_editor.py` tool, which accepts an optional `user_config.yaml`.

mikeroberts3000: If we could set every cvar dynamically, then I would say let's try to completely avoid using an INI string in `user_config.yaml`. But since we need to use the INI string no matter what for `r.RayTracing.Enable=1` or whatever (which can't be set dynamically), then we might as well move `r.RayTracing.SceneCaptures=1` to the INI string also.

mikeroberts3000: Within each of these blocks, do all the `find_console_variable_by_name` calls together, then do all the `set_console_variable_value` calls. It looks cleaner that way.

examples/render_image_path_tracing/run.py
```diff
@@ -0,0 +1,131 @@
...
+        scene_captures_cvar = game.unreal_service.find_console_variable_by_name(console_variable_name="r.RayTracing.SceneCaptures")
+        game.unreal_service.set_console_variable_value(cvar=scene_captures_cvar, value=1)
+        samples_cvar = game.unreal_service.find_console_variable_by_name(console_variable_name="r.PathTracing.SamplesPerPixel")
+        game.unreal_service.set_console_variable_value(cvar=samples_cvar, value=args.num_frames)
+        bounces_cvar = game.unreal_service.find_console_variable_by_name(console_variable_name="r.PathTracing.MaxBounces")
+        game.unreal_service.set_console_variable_value(cvar=bounces_cvar, value=args.bounces)
+
+        # Disable the in-frame path-tracer progress bar (the red line at the bottom) so it can never bake into the
+        # captured image if we happen to read before it clears on its own.
+        progress_display_cvar = game.unreal_service.find_console_variable_by_name(console_variable_name="r.PathTracing.ProgressDisplay")
```
mikeroberts3000: Do you have evidence that this progress bar gets baked into the rendered image? I wouldn't assume that it necessarily gets baked in.

examples/render_image_path_tracing/run.py
```diff
@@ -0,0 +1,131 @@
...
+        game.unreal_service.set_console_variable_value(cvar=progress_display_cvar, value=0)
+
+        # The engine applies the spatial denoiser to the converged image once accumulation reaches
+        # r.PathTracing.SamplesPerPixel. Selecting a denoiser by name requires its plugin to be enabled in
+        # SpearSim.uproject; SpearSim enables "NNEDenoiser" (the engine default) and "OIDN".
+        denoiser_cvar = game.unreal_service.find_console_variable_by_name(console_variable_name="r.PathTracing.Denoiser")
```
mikeroberts3000:
```
        denoiser_cvar = game.unreal_service.find_console_variable_by_name(console_variable_name="r.PathTracing.Denoiser")
        denoiser_name_cvar = game.unreal_service.find_console_variable_by_name(console_variable_name="r.PathTracing.Denoiser.Name")

        if args.denoiser is not None:
            game.unreal_service.set_console_variable_value(cvar=denoiser_cvar, value=1)
            game.unreal_service.set_console_variable_value(cvar=denoiser_name_cvar, value=args.denoiser)
        else:
            game.unreal_service.set_console_variable_value(cvar=denoiser_cvar, value=0)
            game.unreal_service.set_console_variable_value(cvar=denoiser_name_cvar, value="")
```

examples/render_image_path_tracing/run.py
```diff
@@ -0,0 +1,131 @@
...
+        # spawn camera sensor and get the final_tone_curve_hdr component (same as examples/render_image)
+        bp_camera_sensor_uclass = game.unreal_service.load_class(uclass="AActor", name="/SpContent/Blueprints/BP_CameraSensor.BP_CameraSensor_C")
+        bp_camera_sensor = game.unreal_service.spawn_actor(uclass=bp_camera_sensor_uclass)
+        final_tone_curve_hdr_component = game.unreal_service.get_component_by_name(actor=bp_camera_sensor, component_name="DefaultSceneRoot.final_tone_curve_hdr_", uclass="USpSceneCaptureComponent2D")
+
+        # configure the final_tone_curve_hdr component to match the viewport (width, height, FOV, post-processing settings, etc)
+        viewport_desc = game.rendering_service.get_current_viewport_desc()
+        game.rendering_service.align_camera_with_viewport(camera_sensor=bp_camera_sensor, camera_components=final_tone_curve_hdr_component, viewport_desc=viewport_desc, widths=width, heights=height)
+
+        # route this scene-capture view through the path tracer: allow it to use ray tracing and turn on the
```
mikeroberts3000:
```
        # enable path tracing and disable camera imperfections because they amplify noise and produce artifacts if we don't denoise
        final_tone_curve_hdr_component.bUseRayTracingIfEnabled = True
        final_tone_curve_hdr_component.SetShowFlagSettings(InShowFlagSettings=[
            {"ShowFlagName": "PathTracing", "Enabled": True},
            {"ShowFlagName": "CameraImperfections", "Enabled": False}])
```

examples/render_image_path_tracing/run.py
```diff
@@ -0,0 +1,131 @@
...
+        ])
+
+        # need to call initialize_sp_funcs() after calling Initialize() because read_pixels() is registered during Initialize()
+        final_tone_curve_hdr_component.Initialize()
+        final_tone_curve_hdr_component.initialize_sp_funcs()
+
+    with instance.end_frame():
+        pass
+
+    # The path tracer accumulates one additional sample per pixel per rendered frame, exactly like the editor's
```
mikeroberts3000: Great comment.

examples/render_image_path_tracing/user_config.yaml.example
```diff
@@ -0,0 +1,16 @@
...
+# The recommended workflow here for convenience and fast iteration time is LAUNCH_MODE:"none" + EDITOR_LAUNCH_MODE:"game".
+# To run the `run.py`, open the SpearSim project in the UE editor and enter play mode. The render uses the viewport camera.
+
+SPEAR:
+  LAUNCH_MODE: "none"
+
+  INSTANCE:
+    EDITOR_LAUNCH_MODE: "game"
+    # Rendering/denoising can sometimes take a few seconds
+    RPC_CLIENT_INTERNAL_TIMEOUT_SECONDS: 30.0
```
mikeroberts3000: `RPC_CLIENT_INTERNAL_TIMEOUT_SECONDS` is not a config key in SPEAR. The only config key is `CLIENT_INTERNAL_TIMEOUT_SECONDS`.

examples/render_image_path_tracing/user_config.yaml.example
```diff
@@ -0,0 +1,16 @@
+#
+# Copyright (c) 2025 The SPEAR Development Team. Licensed under the MIT License <http://opensource.org/licenses/MIT>.
+# Copyright (c) 2022 Intel. Licensed under the MIT License <http://opensource.org/licenses/MIT>.
+#
+
+# The recommended workflow here for convenience and fast iteration time is LAUNCH_MODE:"none" + EDITOR_LAUNCH_MODE:"game".
+# To run the `run.py`, open the SpearSim project in the UE editor and enter play mode. The render uses the viewport camera.
+
+SPEAR:
+  LAUNCH_MODE: "none"
```
mikeroberts3000: `"game"`

examples/render_image_path_tracing/user_config.yaml.example
```diff
@@ -0,0 +1,16 @@
...
+
+  INSTANCE:
+    EDITOR_LAUNCH_MODE: "game"
```
mikeroberts3000: `EDITOR_LAUNCH_MODE` has no effect here.

examples/render_image_path_tracing/run.py
jakubtomsu: Yes I have actually seen that happen. Sometimes, even when you step the correct number of frames, it would just leak in, similar to the denoiser not being applied even when it should. Not sure why that happens in the first place. The docs say the red bar should never show up but maybe in this case the scene capture component is just wired up exactly like the editor scene viewport, so it shows up anyway.

examples/render_image_path_tracing/user_config.yaml.example
jakubtomsu: Oops that's a hallucination then, thanks for catching that.

examples/render_image_path_tracing/run.py
mikeroberts3000: Ok fair enough. Let's explicitly disable it as you are doing here, although this should be done in the INI string because it doesn't depend on the command-line arguments.

examples/render_image_path_tracing/run.py
```diff
@@ -0,0 +1,131 @@
...
+    # path-tracing viewport, and stops once it reaches r.PathTracing.SamplesPerPixel (set to args.num_frames above).
+    # Moving the camera or anything in the scene invalidates the accumulated samples and restarts from scratch, so we
+    # simply render args.num_frames frames in a row without moving anything, which converges the image on the final
+    # frame. The engine then applies the denoiser (if one was requested) to the converged result.
+    spear.log("Path-traced rendering beginning...")
+
+    for i in range(args.num_frames):
```
mikeroberts3000, line 92: Actually, I think it's too verbose to log every frame. Let's just do this.

```
instance.step(num_frames=args.num_frames, single_step=True)
```

`single_step=True` makes it so UE advances exactly `num_frames` and waits until further notice. Otherwise the SPEAR programming model will intentionally allow UE to advance by some unknown number of frames that is greater than or equal to `num_frames`. I like forcing UE to advance exactly `num_frames` in this example, because we're relying on specific accumulation behavior. The fact that we weren't doing this before might explain some of the seemingly non-deterministic behavior you were seeing.

jakubtomsu: Good idea. I was using the verbose logging as sort of a "progress bar" when testing but I didn't know calling `step(num_frames=...)` had a slightly different behavior

mikeroberts3000 (approved):

PR #320, opened 2026-07-10
Small installation improvements

jakubtomsu:
This PR implements a few tiny changes related to installing SPEAR on Windows.

- ~~Disable Boost ASIO Concepts on MSVC. This prevented me from building on VS2022 even when that should have been fine. This doesn't seem to cause any issues, and it seems nicer to fully disable it rather than just add a flag to ignore the name collision error.~~
- ~~Explicitly tell CMake to use VS2022. Without this flag, it would for some reason use the wrong `cl` *even when I was in the right developer console*.~~
- ~~I added a minimal TL;DR section at the top of `getting_started.md` just to get a simpler overview of the entire installation process. It's pretty much what I've used for multiple SPEAR installations.~~

Of course, it's possible I'm just not familiar enough with some parts of this project or the installation process and there's something obvious I'm missing. Please let me know if that's the case so I can correct/remove some of those changes.

### DIFF
--- a/tools/build_executable_prerequisites.py
+++ b/tools/build_executable_prerequisites.py
@@ -126,7 +126,7 @@ if __name__ == "__main__":
             cmd_prefix + \
             "python " + \
             f'"{os.path.realpath(os.path.join(os.path.dirname(__file__), "build_third_party_libs.py"))}" ' + \
-            f"--third-party-dir  {third_party_dir} "
+            f'--third-party-dir "{third_party_dir}" '
         spear.log("Executing: ", cmd)
         subprocess.run(cmd, shell=True, check=True)
 
--- a/tools/install_python_extension.py
+++ b/tools/install_python_extension.py
@@ -180,7 +180,7 @@ if __name__ == "__main__":
         python_ext_dir = os.path.realpath(os.path.join(os.path.dirname(__file__), "..", "python_ext"))
         so_files = glob.glob(os.path.join(python_ext_dir, "**", "*.so"), recursive=True)
         for so_file in so_files:
-            dsym_cmd = f"dsymutil {so_file} -o {so_file}.dSYM"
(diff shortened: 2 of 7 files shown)


tools/build_executable_prerequisites.py
```diff
@@ -126,7 +126,7 @@
             cmd_prefix + \
             "python " + \
             f'"{os.path.realpath(os.path.join(os.path.dirname(__file__), "build_third_party_libs.py"))}" ' + \
-            f"--third-party-dir  {third_party_dir} "
+            f'--third-party-dir "{third_party_dir}" '
```
mikeroberts3000, line 129: Already fixed.

tools/install_python_extension.py
```diff
@@ -180,7 +180,7 @@
         python_ext_dir = os.path.realpath(os.path.join(os.path.dirname(__file__), "..", "python_ext"))
         so_files = glob.glob(os.path.join(python_ext_dir, "**", "*.so"), recursive=True)
         for so_file in so_files:
-            dsym_cmd = f"dsymutil {so_file} -o {so_file}.dSYM"
+            dsym_cmd = f'dsymutil "{so_file}" -o "{so_file}.dSYM"'
```
mikeroberts3000, line 183: Already fixed.

tools/run_generate_project_files.py
```diff
@@ -50,7 +50,7 @@
     assert len(uprojects) == 1
     uproject = uprojects[0]
 
-    cmd = f'{generate_project_files_script} -project="{uproject}" {generate_project_files_args}'
+    cmd = f'"{generate_project_files_script}" -project="{uproject}" {generate_project_files_args}'
```
mikeroberts3000, line 53: Already fixed.

mikeroberts3000 (approved):
This is great! 😆

mikeroberts3000 (approved):

PR #323, opened 2026-07-20
Monitoring path tracer frame accumulation

jakubtomsu:
This adds a SpFunc to query the internal path tracer frame accumulation counter (previously not available from python), and updates the PT example frame loop to make use of this counter instead of blindly stepping.

The reason why this is even a problem in the first case is not 100% clear to me, but I have run into issues with the frame counting many times before. Even if we call `instance.step(num_frames=num_pt_frames, single_step=True)` with the correct frame count and try really hard to ensure the scene never gets invalidated, it still sometimes breaks. Just this morning I ran the path tracing example with 64 samples and got this (on the last commit and something similar on the one before that):

image_bad

This is clearly completely wrong and nowhere close to the 64 samples, possibly because async loading. Of course, it would be really nice to properly fix the loading too, but this adaptive frame loop at least makes sure the render is correct even in non-perfect conditions.


cpp/unreal_plugins/SpUnrealTypes/Source/SpUnrealTypes/SpSceneCaptureComponent2D.cpp
```diff
@@ -352,6 +353,25 @@ void USpSceneCaptureComponent2D::Initialize()
         }
     });
 
+    SpFuncComponent->registerFunc("get_path_tracing_stats", [this](SpFuncDataBundle& args) -> SpFuncDataBundle {
```
mikeroberts3000: I disagree with this proposed approach. It is brittle (has a hardcoded index 0), will increasingly clutter the capture component implementation as it grows (when we want to return a different combination of rendering stats we will need to constantly add more clutter), doesn't compose well (the capture component is forced to become a wrapper for the lower-level `FSceneViewStateInterface` class in addition to all of its other responsibilities), and is not a good fit for what `SpFunctions` are useful for (passing NumPy arrays to and from Python).

It would be better to create a wrapper class in `SpUnrealTypes` in a separate file called `SpSceneViewStateInterface.h` like this:

```
#include <Kismet/BlueprintFunctionLibrary.h>
#include <Whatever/HeaderIsNeededForFSceneViewStateInterface.h>

#include "SpSceneViewStateInterface.generated.h"

UCLASS()
class USpSceneViewStateInterface : public UBlueprintFunctionLibrary
{
    GENERATED_BODY()
public: 
    UFUNCTION(BlueprintCallable, Category="SPEAR")
    static uint32 GetPathTracingSampleIndex(uint64 ViewState)
    {
        #if RHI_RAYTRACING
            FSceneViewStateInterface* view_state_ptr = reinterpret_cast<FSceneViewStateInterface*>(ViewState);
            SP_ASSERT(view_state_ptr);
            return view_state_ptr->GetPathTracingSampleIndex();
        #else
            return 0;
        #endif
    }

    UFUNCTION(BlueprintCallable, Category="SPEAR")
    static uint32 GetPathTracingSampleCount(uint64 ViewState)
    {
        #if RHI_RAYTRACING
            FSceneViewStateInterface* view_state_ptr = reinterpret_cast<FSceneViewStateInterface*>(ViewState);
            SP_ASSERT(view_state_ptr);
            return view_state_ptr->GetPathTracingSampleCount();
        #else
            return 0;
        #endif
    }
};
```

I like this approach better because it doesn't clutter the capture component and is completely out of the way (no other C++ source file even needs to reference it). It can grow organically as we want to expose more rendering stats because we can just add more functions to `USpSceneViewStateInterface` (not a big deal because it is so out of the way). Many of the classes in `SpUnrealTypes` are wrappers that follow this pattern.

To support my proposed approach, I've already created a `GetViewStates` `UFUNCTION` that returns an array of pointers cast to `uint64` (necessary because `FSceneViewStateInterface` is not visible to the reflection system and is therefore not allowed in `UFUNCTION` signatures).

From Python, a user could do something like:

```
sp_scene_view_state_interface = game.get_unreal_object(uclass="USpSceneViewStateInterface")

# ...

view_states = component.GetViewStates()
if len(view_states) > 0: # 0 index assumption lives in user code
    view_state = view_states[0]
    sample_index = sp_scene_view_state_interface.GetPathTracingSampleIndex(ViewState=view_state)
    sample_count = sp_scene_view_state_interface.GetPathTracingSampleCount(ViewState=view_state)
else:
    sample_index = 0
    sample_count = 0
```

examples/render_image_path_tracer/run.py
```diff
@@ -94,14 +98,33 @@
...
+    # ensure the image has been fully rendered. This way we avoid potential loading issues.
+    # The engine then applies the denoiser (if one was requested) to the converged result.
     spear.log("Path-traced rendering beginning...")
-    for i in range(args.num_frames):
-        spear.log(f"Rendering frame {i:04d}...")
-        instance.step(single_step=True)
+    max_num_frames = args.num_frames*2
+    sample_index = 0
+    sample_count = 0
+    for i in range(max_num_frames):
```
mikeroberts3000: There is no harm in allowing the path tracer to run for a few extra frames. On the final frame, the denoiser will run when it is supposed to. On subsequent frames, the path tracer will stop accumulating samples but the final image will still be available to retrieve later. Is the path tracer not behaving like this for you? Are you finding that the path tracer doesn't always accumulate a new sample during each UE frame (i.e., there are sometimes UE frames that don't advance the path tracer).

cpp/unreal_plugins/SpUnrealTypes/Source/SpUnrealTypes/SpSceneCaptureComponent2D.cpp
jakubtomsu: Right, the separate UFUNCTION wrapper does indeed look nice. Thank you for the suggestion, I used SpFunc as it was the easiest/quickest way to get it working.

examples/render_image_path_tracer/run.py
jakubtomsu: Yes, for some reason even many extra frames aren't enough. I guess the engine just stalls, I don't know why. But the render I attached above was after a fresh engine boot. The progress bar shows it has barely started, even though the SPEAR app literally stepped over 64 frames...

I just find this "adaptive" approach much nicer, and also from the developer's perspective I have insight into the actual internal accumulation counter state, instead of guessing if it's off-by-one somehow.

cpp/unreal_plugins/SpUnrealTypes/Source/SpUnrealTypes/SpSceneCaptureComponent2D.cpp
jakubtomsu: I will update the PR with your code and run my tests, unless you want to do the changes yourself

mikeroberts3000: Please go ahead 😄

examples/render_image_path_tracer/run.py
mikeroberts3000: Yes agreed. Also is that progress bar from `read_pixels` or a screenshot? If it is from `read_pixels`, I guess we really do need to disable progress bar drawing explicitly. (I couldn't reproduce that behavior on my Windows machine for some reason so I took it out.)

jakubtomsu: It really is from `read_pixels`, not a screenshot or something. It is quite weird. I *think* it's just because we're rendering in interactive mode, not the offline render mode the MRQ uses (that's actually a separate thing, the interactive mode was the easiest to set up, but it might be worth it looking into overriding the offline mode flag from C++ at some point to see what feature that unlocks. I haven't looked into it yet)

mikeroberts3000: Why do you suppose I can't reproduce it though? When I run the standard PT example as it exists in `main`, I can't reproduce with `LAUNCH_MODE == "none"` in the editor or `LAUNCH_MODE == "game"` in a standalone executable.

jakubtomsu: That is a good question and I'm not sure. I think it might be because my laptop is much slower. This happens especially when scenes are loaded cold from disk. But it doesn't happen very frequently to me as well so I have a hard time investigating it. Usually it happens in the first render and then in later renders everything is fine.

cpp/unreal_plugins/SpUnrealTypes/Source/SpUnrealTypes/SpSceneViewStateInterface.h
```diff
@@ -0,0 +1,43 @@
...
+class USpSceneViewStateInterface : public UBlueprintFunctionLibrary
+{
+    GENERATED_BODY()
+public: 
+    UFUNCTION(Category="SPEAR") // uint64 is not supported for BlueprintCallable
+    static uint32 GetPathTracingSampleIndex(uint64 ViewState)
+    {
+        #if RHI_RAYTRACING
+            FSceneViewStateInterface* view_state_ptr = reinterpret_cast<FSceneViewStateInterface*>(ViewState);
+            if (view_state_ptr) {
```
mikeroberts3000:
```
SP_ASSERT(view_state_ptr);
```

cpp/unreal_plugins/SpUnrealTypes/Source/SpUnrealTypes/SpSceneViewStateInterface.h
```diff
@@ -0,0 +1,43 @@
...
+                return view_state_ptr->GetPathTracingSampleIndex();
+            }
+        #endif
```
mikeroberts3000, line 32:
```
#else
    return 0;
#endif
```

cpp/unreal_plugins/SpUnrealTypes/Source/SpUnrealTypes/SpSceneViewStateInterface.h
```diff
@@ -0,0 +1,43 @@
...
+
+        return 0;
+    }
+
+    UFUNCTION(Category="SPEAR") // uint64 is not supported for BlueprintCallable
+    static uint32 GetPathTracingSampleCount(uint64 ViewState)
```
mikeroberts3000, line 36: See above.

examples/render_image_path_tracer/run.py
mikeroberts3000: This is a bug. `BP_CameraSensorPathTracer` is needed because the example code isn't setting the show flags for path tracing any more. That only happens in `BP_CameraSensorPathTracer`. So we need to use `BP_CameraSensorPathTracer`.

examples/render_image_path_tracer/run.py
```diff
@@ -115,11 +117,15 @@
                 final_tone_curve_hdr_component.RequestPathTracerReset()
 
         with instance.end_frame(single_step=True):
-            path_tracing_stats_data_bundle = final_tone_curve_hdr_component.get_path_tracing_stats()
+            view_states = final_tone_curve_hdr_component.GetViewStates()
+            if len(view_states) > 0:
```
mikeroberts3000:
```
assert len(view_states) > 0
view_state = view_states[0]
sample_index = sp_scene_view_state_interface.GetPathTracingSampleIndex(ViewState=view_state)
sample_count = sp_scene_view_state_interface.GetPathTracingSampleCount(ViewState=view_state)
```

examples/render_image_path_tracer/run.py
mikeroberts3000: Ok no worries. Then let's add:

```
game.console_service.set(name="r.PathTracing.ShowProgress", value=0)
```

Or whatever directly above where we're setting `r.PathTracing.SamplesPerPixel`.

examples/render_image_path_tracer/run.py
```diff
@@ -7,6 +7,7 @@
 
 import argparse
 import cv2
+import json
```
mikeroberts3000: Not needed.

examples/render_image_path_tracer/run.py
```diff
@@ -76,10 +81,20 @@
         visualize_func = lambda data : data[:,:,[2,1,0]] # BGRA -> RGB
         # visualize_func = lambda data : data
 
+        # enable path tracing and disable camera imperfections because they amplify noise and produce artifacts if we don't denoise
```
mikeroberts3000: Adjusting show flags not needed because now it happens in the definition of the camera sensor.

examples/render_image_path_tracer/run.py
```diff
@@ -54,10 +55,14 @@
         # force high-res textures for captured images
         game.console_service.set(name="r.Streaming.FullyLoadUsedTextures", value=1)
 
+        # workaround for nanite rebuilds invalidating the path tracer state
```
mikeroberts3000:
```
        # When path tracing begins, Unreal begins asynchronously building various Nanite data structures, and
        # as these data structures finish building, they internally reset the progress of the path tracer.
        # Unfortunately Unreal doesn't expose any way to check if these asynchronous build tasks are complete,
        # so we proceed with a heuristic approach that attempts to front-load the build tasks as much as
        # possible, and then we render a conservative number of warm-up frames to ensure the build tasks are
        # complete before we start rendering a final image.

        game.console_service.set(name="r.RayTracing.Nanite.ForceUpdateVisible", value=1)                     # force all Nanite builds to be scheduled on a single frame
        game.console_service.set(name="r.RayTracing.Nanite.MaxBuiltPrimitivesPerFrame", value=256*1024*1024) # set the max number of built Nanite primitives per frame to be very large (default is 8*1024*1024)
        game.console_service.set(name="r.RayTracing.Nanite.MaxStagingBufferSizeMB", value=2048)              # set the size of the Nanite staging buffer to be large (default is 1024)
```

examples/render_image_path_tracer/run.py
```diff
@@ -76,10 +81,20 @@
         visualize_func = lambda data : data[:,:,[2,1,0]] # BGRA -> RGB
         # visualize_func = lambda data : data
 
+        # enable path tracing and disable camera imperfections because they amplify noise and produce artifacts if we don't denoise
+        final_tone_curve_hdr_component.SetShowFlagSettings(InShowFlagSettings=[
+            {"ShowFlagName": "PathTracing", "Enabled": True},
+            {"ShowFlagName": "CameraImperfections", "Enabled": False}])
+
+        # Required for RequestPathTracerReset() (called below) to have any effect
```
mikeroberts3000: Setting `bUseSceneViewExtension` explicitly is not needed any more. I changed the camera sensor to auto-detect if a scene view extension is needed based on the show flags and various other things.

examples/render_image_path_tracer/run.py
```diff
@@ -89,14 +104,32 @@
     # The path tracer accumulates one sample per pixel per rendered frame, exactly like the editor's
     # path-tracing viewport, and stops once it reaches r.PathTracing.SamplesPerPixel (set to args.num_frames
     # above). Moving the camera or anything in the scene invalidates the accumulated samples and restarts
-    # from scratch, so we simply render args.num_frames frames in a row without moving anything, which
```
mikeroberts3000, line 92:
```
    # inserting an extra frame or two can fix occasional render-to-texture initialization issues (advances a minimum of 3 frames)
    game.async_loading_service.wait_for_engine_idle()

    spear.log("Rendering warm-up frames...")

    # advance an extra few frames to give Nanite a chance to finish building its data structures
    instance.step(num_frames=args.num_warmup_frames)

    # force Nanite to stop building its data structures so it can't invalidate the path tracer any more
    with instance.begin_frame():
        game.console_service.set(name="r.RayTracing.Nanite.Update", value=0)
    with instance.end_frame():
        pass
   
    spear.log("Finished rendering warm-up frames.")

    # The path tracer accumulates one sample per pixel per rendered frame, exactly like the editor's
    # path-tracing viewport, and stops once it reaches r.PathTracing.SamplesPerPixel (set to args.num_frames
    # above). Moving the camera or anything in the scene invalidates the accumulated samples and restarts
    # from scratch. We poll the internal path tracer accumulation counter to ensure that the image has been
    # fully rendered. This way we avoid potential loading issues. The engine then applies the denoiser (if
    # one was requested) to the converged result.

    spear.log("Path-traced rendering beginning...")

    for i in range(args.num_frames):

        with instance.begin_frame():
            # Explicitly reset the path tracer's accumulated samples on the first frame. Nothing moves
            # in this example, but frames rendered above (e.g. during wait_for_engine_idle()) already accumulated
            # samples against this component's persistent view state, so without this reset, sample_index would
            # start ahead of 0 below.
            if i == 0:
                final_tone_curve_hdr_component.RequestPathTracerReset()

        with instance.end_frame(single_step=True):
            view_states = final_tone_curve_hdr_component.GetViewStates()
            assert len(view_states) == 1
            view_state = view_states[0]
            sample_index = sp_scene_view_state_interface.GetPathTracingSampleIndex(ViewState=view_state)
            assert sample_index == i + 1

    spear.log("Path-traced rendering finished.")
```

Set a reasonable default value for `args.num_warmup_frames` that works reliably for you.

cpp/unreal_plugins/SpUnrealTypes/Source/SpUnrealTypes/SpSceneViewStateInterface.h
```diff
@@ -0,0 +1,44 @@
+//
+// Copyright (c) 2025 The SPEAR Development Team. Licensed under the MIT License <http://opensource.org/licenses/MIT>.
+// Copyright (c) 2022 Intel. Licensed under the MIT License <http://opensource.org/licenses/MIT>.
+//
+
+#pragma once
+
+#include <Kismet/BlueprintFunctionLibrary.h>
+#include <SceneManagement.h> // FSceneViewStateInterface
```
mikeroberts3000, line 9:
```
#include <SceneManagement.h> // FSceneViewStateInterface
#include <RenderingThread.h> // FlushRenderingCommands
```

cpp/unreal_plugins/SpUnrealTypes/Source/SpUnrealTypes/SpSceneViewStateInterface.h
```diff
@@ -0,0 +1,44 @@
...
+UCLASS()
+class USpSceneViewStateInterface : public UBlueprintFunctionLibrary
+{
+    GENERATED_BODY()
+public: 
+    UFUNCTION(Category="SPEAR") // uint64 is not supported for BlueprintCallable
+    static uint32 GetPathTracingSampleIndex(uint64 ViewState)
+    {
+        #if RHI_RAYTRACING
+            FSceneViewStateInterface* view_state_ptr = reinterpret_cast<FSceneViewStateInterface*>(ViewState);
```
mikeroberts3000, line 26:
```
        #if RHI_RAYTRACING
            FlushRenderingCommands(); // force rendering thread to be fully up-to-date before querying GetPathTracingSampleIndex()
            FSceneViewStateInterface* view_state_ptr = reinterpret_cast<FSceneViewStateInterface*>(ViewState);
```

cpp/unreal_plugins/SpUnrealTypes/Source/SpUnrealTypes/SpSceneViewStateInterface.h
```diff
@@ -0,0 +1,44 @@
...
+        #else
+            return 0;
+        #endif
+    }
+
+    UFUNCTION(Category="SPEAR") // uint64 is not supported for BlueprintCallable
+    static uint32 GetPathTracingSampleCount(uint64 ViewState)
+    {
+        #if RHI_RAYTRACING
+            FSceneViewStateInterface* view_state_ptr = reinterpret_cast<FSceneViewStateInterface*>(ViewState);
```
mikeroberts3000, line 40:
```
        #if RHI_RAYTRACING
            FlushRenderingCommands(); // force rendering thread to be fully up-to-date before querying GetPathTracingSampleCount()
            FSceneViewStateInterface* view_state_ptr = reinterpret_cast<FSceneViewStateInterface*>(ViewState);
```

mikeroberts3000 (approved):
Great.

PR #324, opened 2026-07-20
Minimal RenderDoc + SPEAR usage example

jakubtomsu:
This is a very simple example, especially because UE sets-up the user-defined RenderDoc capture range for us. All we have to do is to just tell it to capture a specific SceneCaptureComponent render. Apart from that, it's just about setting up the ini flags correctly.

But importantly this example works out-of-the-box once you install RenderDoc.


examples/renderdoc_capture/user_config.yaml.example
```diff
@@ -0,0 +1,24 @@
...
+  INSTANCE:
+    EDITOR_EXECUTABLE: "C:/Program Files/Epic Games/UE_5.5/Engine/Binaries/Win64/UnrealEditor.exe"
+    EDITOR_UPROJECT: "C:/repos/spear/cpp/unreal_projects/SpearSim/SpearSim.uproject"
+    EDITOR_LAUNCH_MODE: "game"
+    CLIENT_INTERNAL_TIMEOUT_SECONDS: 60.0
+
+SP_CORE:
+  # Adjust renderdoc.BinaryPath to your RenderDoc install.
+  OVERRIDE_CONFIG_ENGINE_INI: True
+  CONFIG_ENGINE_INI_STRING: |
```
mikeroberts3000: Does this example require enabling some kind of `RenderDoc` plugin? If there is a plugin, is there any performance penalty associated with enabling the plugin and not capturing anything? In other words, is there any downside to leaving it enabled all the time?

examples/renderdoc_capture/user_config.yaml.example
```diff
@@ -0,0 +1,24 @@
...
+    [/Script/RenderDocPlugin.RenderDocPluginSettings]
+    renderdoc.AutoAttach=True
+    renderdoc.BinaryPath=C:\Program Files\RenderDoc
+    renderdoc.CaptureAllActivity=False
+    renderdoc.ReferenceAllResources=False
+    renderdoc.CaptureCallstacks=False
```
mikeroberts3000: `CaptureCallstacks` seems like it would be useful. What made you choose these INI values? Are they default values?

examples/renderdoc_capture/run.py
```diff
@@ -0,0 +1,70 @@
...
+    with instance.end_frame():
+        pass
+
+    # let temporal anti-aliasing etc accumulate additional information across multiple frames, and inserting an
+    # extra frame or two can fix occasional render-to-texture initialization issues (advances a minimum of 3 frames)
+    game.async_loading_service.wait_for_engine_idle()
+
+    # set-up the scene capture component to do an isolated RenderDoc GPU capture
+    with instance.begin_frame():
+        final_tone_curve_hdr_component.bCaptureGpuNextRender = True
```
mikeroberts3000: This is the only unique line of code relative to `render_image`. So I'm thinking that there could be an optional argument in `render_image` called `--capture` that would set `bCaptureGpuNextRender = True`.

examples/renderdoc_capture/user_config.yaml.example
```diff
@@ -0,0 +1,24 @@
...
+
+SPEAR:
+  LAUNCH_MODE: "editor"
+  INSTANCE:
+    EDITOR_EXECUTABLE: "C:/Program Files/Epic Games/UE_5.5/Engine/Binaries/Win64/UnrealEditor.exe"
+    EDITOR_UPROJECT: "C:/repos/spear/cpp/unreal_projects/SpearSim/SpearSim.uproject"
+    EDITOR_LAUNCH_MODE: "game"
+    CLIENT_INTERNAL_TIMEOUT_SECONDS: 60.0
+
+SP_CORE:
```
mikeroberts3000: This is so similar to the `render_image` example. It really only differs by a single line of code. I'm thinking that this config block could just be a comment in `render_image`.

examples/renderdoc_capture/user_config.yaml.example
jakubtomsu: Oh this is a good point. The RenderDoc plugin is built-in, but I haven't checked if it's enabled or disabled by default.

Running with renderdoc enabled all the time is not an issue - it pretty much no-ops until the user/program triggers a GPU capture. At that point the frame render slows down a lot (10x or more) because it needs to clone all the resources and capture lots of GPU and graphics API state. But that's only for a single frame. So RenderDoc is totally free when not capturing anything.

examples/renderdoc_capture/run.py
jakubtomsu: Agreed, I also got that idea, I just wasn't sure if we wanted a standalone example or no, especially since the INI flags make a big difference.

examples/renderdoc_capture/user_config.yaml.example
jakubtomsu: Yes call stacks can be useful but only in special cases when you actually need to go and look at the rendering code.

In my experience, most of the time you just want to do a frame capture and see the draw commands. UE already submits a LOT of labels/sections to help structuring the render passes. That's the important part for navigating the capture in RenderDoc, call stacks are unnecessary.

Here you can see how the frame data is structured into a hierarchy timeline even without callstacks:

image

jakubtomsu: Regarding the other flags, I pretty much turn everything off because it's *extra* information. The base capture already has everything I need, and it's very big (~7 GB for a 1080p PT capture for example), so I'm just trying to trim it down by default. It can be easily turned on when actually necessary ofc.

examples/renderdoc_capture/user_config.yaml.example
jakubtomsu: Makes sense. I will look into merging those two, it looks like there should be zero issues.

examples/renderdoc_capture/user_config.yaml.example
mikeroberts3000: Update: enabling `XcodeGPUDebuggerPlugin` costs about 0.5 ms per frame on my macOS laptop even when I'm not capturing anything. `PixWinPlugin` and `RenderDocPlugin` might be more lightweight, but my vote is to include {`PixWinPlugin`, `RenderDocPlugin`, `XcodeGPUDebuggerPlugin`} in `SpearSim.uproject` but have them all set to `Enabled: False` for consistency.

mikeroberts3000: Ah, never mind, `PixWinPlugin` and `RenderDocPlugin` are enabled by default. So we don't need to do anything extra there. `XcodeGPUDebuggerPlugin` is not enabled by default (probably because it makes rendering a bit slower even when it's not doing anything), so I will add it to `SpearSim.uproject` but I will set it to be disabled by default.

examples/renderdoc_capture/user_config.yaml.example
mikeroberts3000: Reasonable.

mikeroberts3000: Where did the INI values come from though? How did you know what to set? Did you configure the options in the project settings and see which changes were written to DefaultEngine.ini or something?

jakubtomsu: Yes, I used the project settings as a reference. It seems worth it having the flags specified in the user config instead of just telling users to configure it themselves

mikeroberts3000 (approved):
This is great!

PR #325, opened 2026-07-21
Basic github workflows

jakubtomsu:
This adds basic CI to check python code with a linter, verify spear python installation on Windows, Ubuntu and MacOS, and test `spear_ext` compiles on Windows and MacOS. This is quite experimental but could be quite helpful especially in catching stale code bugs and platform specific issues.

The linter has already discovered a few tiny errors, so the fixes are a part of this PR as well.

Tested here: https://github.com/jakubtomsu/spear/pull/1


examples/getting_started_notebook/getting_started.ipynb
```diff
@@ -1870,7 +1870,7 @@
     "\n",
     "    # get the result from our previous non-blocking call\n",
     "    location = future.get()\n",
-    "\n",
+    "\n"
```
mikeroberts3000, line 1873: This is how Jupyter saves the file. This file isn't really a source file in the same sense as the others. We should exclude it from being modified as part of this automated testing.

python/spear/utils/editor_utils.py
```diff
@@ -383,7 +383,7 @@ def to_unreal_transform_from_numpy_transform(numpy_transform):
     return to_unreal_transform_from_spear_transform(spear_transform=spear.math.to_spear_transform_from_numpy_transform(numpy_transform=numpy_transform))
 
 def to_unreal_transform_from_numpy_matrix(matrix):
```
mikeroberts3000: Fix this in the opposite way: rename `matrix` -> `numpy_matrix`, not the other way around.

mikeroberts3000:
I separated the bug fixes from the workflow itself and already fixed the bugs and checked in fixes separately. I'm going to examine the workflow itself in more detail and try to understand it a bit better before approving.

mikeroberts3000:
Also, the JSON error that was identified by this git workflow actually was a problem in the JSON. I previously made a comment that we should exclude the notebook from being checked because I thought the workflow was reporting a false positive. But actually the git workflow was correct and my comment was incorrect. I fixed the JSON in `main`.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,105 @@
...
+
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.11"
+
+      - name: Install the spear Python package
+        run: pip install -e "python[examples]"
+
+      - name: Import spear
+        run: python -c "import spear; spear.log('spear imported successfully, __can_import_spear_ext__ = ', spear.__can_import_spear_ext__)"
```
mikeroberts3000:
```
       run: python -c "import spear; spear.log('spear imported successfully (__can_import_spear_ext__ = ', spear.__can_import_spear_ext__, ')')"
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,105 @@
...
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.11"
+
+      - name: Check Python syntax
```
mikeroberts3000: Isn't this redundant with the `ruff` step? What errors would be caught with `compileall` that wouldn't be caught with `ruff`?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,105 @@
...
+        run: python -m compileall -q python tools pipeline editor examples python_ext/python
+
+      - name: Lint with ruff
```
mikeroberts3000:
```
      - name: Lint Python source code
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,105 @@
...
+      - "LICENSE.txt"
+      - "ACKNOWLEDGMENTS.txt"
+  pull_request:
+    paths-ignore: *docs_paths
+  workflow_dispatch:
+
+jobs:
+
+  # Pure static analysis. Results don't depend on OS, so this only needs to run once.
+  lint:
```
mikeroberts3000:
```
  lint-python:
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,105 @@
...
+      - name: Check Python syntax
+        run: python -m compileall -q python tools pipeline editor examples python_ext/python
+
+      - name: Lint with ruff
+        run: |
+          pip install ruff==0.15.22
+          ruff check --select F --ignore F401,F403,F541,F841 python tools pipeline editor examples python_ext/python
+
+  # Verify the spear Python package actually installs and imports on each platform.
+  install:
```
mikeroberts3000:
```
  install-python:
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,105 @@
...
+        with:
+          arch: x64
+
+      - if: runner.os == 'macOS'
+        run: sudo xcode-select -s /Applications/Xcode_26.5.app
+
+      - name: Install the spear Python package
+        run: conda run --no-capture-output -n spear-env pip install -e python
+
+      - name: Build third-party libraries
```
mikeroberts3000, line 101: What is the purpose of `conda run -n spear-env` and the `activate-environment: spear-env` step above? Are both really necessary?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,105 @@
...
+
+      - name: Import spear
+        run: python -c "import spear; spear.log('spear imported successfully, __can_import_spear_ext__ = ', spear.__can_import_spear_ext__)"
+
+  # Compile the spear_ext Python extension module and its vendored third-party libraries
+  # (boost, rpclib, yaml-cpp) using the system compiler. This intentionally excludes Linux,
+  # because our build scripts source clang/libc++ from an Unreal Engine install on Linux.
+  #
+  # We pin windows-2022 rather than windows-latest because MSVC v143 (v14.44-17.14) is required.
+  compile-spear-ext:
```
mikeroberts3000:
```
  install-python-ext:
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,105 @@
...
+        with:
+          python-version: "3.11"
+
+      - name: Install the spear Python package
+        run: pip install -e "python[examples]"
+
+      - name: Import spear
+        run: python -c "import spear; spear.log('spear imported successfully, __can_import_spear_ext__ = ', spear.__can_import_spear_ext__)"
+
+  # Compile the spear_ext Python extension module and its vendored third-party libraries
```
mikeroberts3000:
```
  # Compile the spear_ext Python extension module and its vendored third-party libraries
  # (boost, rpclib, yaml-cpp) using the system compiler. We intentionally exclude Linux
  # because our build scripts source clang/libc++ from an Unreal Engine install on Linux.
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,105 @@
...
+        run: |
+          pip install ruff==0.15.22
+          ruff check --select F --ignore F401,F403,F541,F841 python tools pipeline editor examples python_ext/python
+
+  # Verify the spear Python package actually installs and imports on each platform.
+  install:
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, windows-latest, macos-latest]
```
mikeroberts3000: I'd rather pin specific versions for all OSes here. It will be easy for `latest` to break later. Will `/Applications/Xcode_26.5.app` even run on `macos-latest` a year from now?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,105 @@
...
+  # Compile the spear_ext Python extension module and its vendored third-party libraries
+  # (boost, rpclib, yaml-cpp) using the system compiler. This intentionally excludes Linux,
+  # because our build scripts source clang/libc++ from an Unreal Engine install on Linux.
+  #
+  # We pin windows-2022 rather than windows-latest because MSVC v143 (v14.44-17.14) is required.
+  compile-spear-ext:
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [windows-2022, macos-latest]
```
mikeroberts3000: It is better to pin all OS versions.

.github/workflows/ci.yml
jakubtomsu: I think there's no reason not to run both. Both catch compile errors, but [ruff has a custom parser](https://github.com/astral-sh/ruff/tree/main/crates/ruff_python_parser) so it's not always 100% equal (though the compatibility should be very good). Also `compileall` should do more than pure syntactic/semantic analysis, it also runs the bytecode generator which can trigger some errors apparently. On the other hand `ruff` catches a lot of potential issues the reguler compiler doesn't check for.

.github/workflows/ci.yml
jakubtomsu: I've had issues with the windows shell not preserving environment variables/context across steps. But I'm not sure that applies anymore, I will try to run it again without that.

.github/workflows/ci.yml
mikeroberts3000: Fair, but having both adds complexity to the codebase. So I only want to do both if there isn't a containment relationship between the two checks. What would be an error that is caught by `compileall` but not `ruff`? I asked this before and you just said "some errors apparently". What is one such error?

.github/workflows/ci.yml
mikeroberts3000: If we need `conda run -n spear-env` on Windows, then my thinking is that we should just use that everywhere, and shouldn't also use `activate-environment`.

jakubtomsu: Turns out something else must have been wrong when I ran into the problems in the past, now it runs correctly without the `conda run ...` even on windows.

mikeroberts3000 (approved):
Great.

PR #326, opened 2026-07-22
Added guard to SpCore's callFunction against stale/dangling UObject handles

soravux:
UnrealService::callFunction could SIGSEGV when a Python client called a reflected UFUNCTION on a UObject (or UFunction) that the engine had garbage-collected or destroyed since the handle was obtained. The raw pointer is non-null (so it passes the existing SP_ASSERT(uobject) / SP_ASSERT(ufunction)), but points at freed memory, and the crash lands deep in the reflected call with no indication of which function or object was at fault.

The PR adds an IsValidLowLevelFast() check on both ufunction and uobject immediately after the existing asserts to validate a possibly-garbage pointer (vtable + GUObjectArray index) without dereferencing user memory, so a stale handle now fails with a clear, catchable error that logs the offending function name instead of a segfault.

### DIFF
--- a/cpp/unreal_plugins/SpCore/Source/SpCore/UnrealUtils.cpp
+++ b/cpp/unreal_plugins/SpCore/Source/SpCore/UnrealUtils.cpp
@@ -166,6 +166,18 @@ std::map<std::string, SpPropertyValue> UnrealUtils::callFunction(const UWorld* w
     SP_ASSERT(world);
     SP_ASSERT(uobject);
     SP_ASSERT(ufunction);
+
+    // Ensure UFunction/UObject are not garbage-collected.
+    if (!ufunction->IsValidLowLevelFast()) {
+        SP_LOG("ERROR: callFunction: ufunction is a stale/invalid handle; refusing to call.");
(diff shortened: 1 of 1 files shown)


mikeroberts3000:
I think it would be better to do this validation in a single catch-all place that handles all pointers coming across the Python / C++ boundary, instead of only handling `callFunction`.

@jakubtomsu, do you want to implement this? You could create a new specialization for `Service::toPtr` using the `CObject` concept. The specialized logic should be: if the input `uint64_t` is zero, then `reinterpret_cast` it to a pointer and don't do anything else. If it is non-zero, then assert `IsValidLowLevelFast` and `IsValidLowLevel` and `IsValid` (all checks are necessary).

jakubtomsu:
@mikeroberts3000 yes absolutely, I will look into it

mikeroberts3000:
As part of a larger change, I ended up fixed this issue in `main` in a more general way than the one proposed in this PR. I am now validating all `UObject` pointers that get passed into C++ code from Python, so I will catch all errors that get caught by this PR and many others. Thanks for reporting 🥳

PR #327, opened 2026-07-24
UST captures

jakubtomsu:
This implements two alternative camera sensor blueprints after Mike's recent breakthrough change that added UST support.

Changes involved:
- Generator for PPM instances. Fully duplicating the materials would be wasteful, especially since the instances have a way to override the UST name field. Nothing else needs to change on the materials (the resolution divisor can stay at 0!) so this lets us keep the original PPM as a source of truth and re-generate whatever we need. That said I am not sure whether these instances should be checked into source control or generated dynamically.
- New generators for UST camera sensor blueprints. The number of capture components reduced dramatically, because we get most of the buffers on every render for free. I haven't checked everything 100% but the current setup seems to be able to produce all modalities the old one could, except object IDs since those require some kind of a proxy (???)
- Small C++ tweak to let us use both material instances and base materials through `UMaterialInterface`

The example is a little sloppy - I haven't cleaned it up yet as I'm still experimenting. It dumps all the channels supported by a specific capture component, so for now it's mostly useful for debugging and tweaking the component config.


jakubtomsu:
Example output:

image

mikeroberts3000:
Great! Very nice approach to create material instances instead of duplicating the materials. I'm not worried about the wasted disk space, but this approach will probably be easier to keep the pairs in sync in case we want to change the base materials.

cpp/unreal_plugins/SpUnrealTypes/Source/SpUnrealTypes/SpSceneCaptureComponent2D.h
```diff
@@ -50,6 +50,7 @@ class FRDGBuilder;
 class FRHICommandListImmediate;
 class FRHITexture;
 class UMaterial;
```
mikeroberts3000: Remove forward declare because it isn't used.

```
class UMaterial;
```

editor/create_asset_bp_camera_sensor_path_tracer_ust.py
```diff
@@ -0,0 +1,213 @@
...
+# reconstruction does not survive the path tracer. The enabled set (UserSceneTextureNames) is left empty here;
+# consumers select buffers at runtime.
+user_scene_texture_material_names = \
+[
+    "PathTracingAlbedo",
+    "PathTracingDenoisedRadiance",
+    "PathTracingNormal",
+    "PathTracingRadiance",
+    "PathTracingVariance",
+    "SceneDepth"
```
mikeroberts3000, line 30: For simplicity, components that enable the path tracer should only enable the `PathTracing*` USTs. So `SceneDepth` should be removed here. Or, if you really want to enable `SceneDepth` here, then you should enable all valid USTs. It is weird to single out `SceneDepth` here.

editor/create_asset_bp_camera_sensor_path_tracer_ust.py
```diff
@@ -0,0 +1,213 @@
...
+            "fov_angle": fov_angle,
+            "num_channels_per_pixel": 4,
+            "channel_data_type": unreal.SpArrayDataType.U_INT8,
+            "capture_source": unreal.SceneCaptureSource.SCS_FINAL_TONE_CURVE_HDR,
+            "show_flag_settings": engine_show_flag_settings["with_path_tracing"],
+            "override_texture_render_target_format": True,
+            "texture_render_target_format": unreal.TextureRenderTargetFormat.RTF_RGBA8_SRGB
+        },
+        {
+            "name": "lighting_only_",
```
mikeroberts3000, line 116: We need a new naming convention here. The name of each components should map 1:1 with distinct rendering passes rather than specific PPMs as we have done in previous camera sensor implementations. Suggested names:

```
final_tone_curve_hdr_
disable_all_
lighting_only_
lighting_only_with_path_tracing_
specular_only_with_path_tracing_
with_lighting_with_path_tracing_
```

Using an appropriate combination of USTs from these passes, it is possible to compute all results in `rendered_image_hypersim` but using the path tracer.

editor/create_asset_bp_camera_sensor_ust.py
```diff
@@ -0,0 +1,253 @@
...
+
+        sp_scene_capture_component_2d.set_editor_property(name="width", value=component_desc["width"])
+        sp_scene_capture_component_2d.set_editor_property(name="height", value=component_desc["height"])
+        sp_scene_capture_component_2d.set_editor_property(name="num_channels_per_pixel", value=component_desc["num_channels_per_pixel"])
+        sp_scene_capture_component_2d.set_editor_property(name="channel_data_type", value=component_desc["channel_data_type"])
+        sp_scene_capture_component_2d.set_editor_property(name="capture_source", value=component_desc["capture_source"])
+
+        # SpSceneCaptureComponent2D properties (optional)
+
+        if "use_scene_view_extension" in component_desc:
```
mikeroberts3000: `use_scene_view_extension` is auto-detected now so this is not needed.

editor/create_asset_bp_camera_sensor_ust.py
```diff
@@ -0,0 +1,253 @@
...
+fov_angle = 90.0
+
+# The "universe" of UserSceneTexture buffers exposed on every capture component. Each name maps to the
+# MI_PPM_<name>_UST material instance (see create_asset_ppm_ust_instances.py), whose post-process output is
+# redirected to a UserSceneTexture of that name. We apply the universe to every component because different
+# components render different passes (e.g. lighting-only), and a buffer can only be produced by the pass that
+# renders it. Depth (SpDepthMeters) is included so it is available from every component. The enabled set
+# (UserSceneTextureNames) is left empty here; consumers select buffers at runtime.
+user_scene_texture_material_names = \
+[
```
mikeroberts3000, line 24: Good.

editor/create_asset_bp_camera_sensor_ust.py
```diff
@@ -0,0 +1,253 @@
...
+    unreal.EngineShowFlagsSetting(show_flag_name="TemporalAA", enabled=True)
+]
+
+#
+# final_tone_curve_hdr without temporal anti-aliasing. TAA only accumulates the scene color, so the beauty is
+# not anti-aliased here, but the GBuffer-derived aux buffers (normals, depth, etc.) are captured without the
+# sub-pixel jitter that TAA introduces on every frame.
+#
+
+engine_show_flag_settings["final_tone_curve_hdr_no_taa"] = []
```
mikeroberts3000: Let's call this `with_lighting` (and move the definition of it to be after `lighting_only`) for consistency with the non-UST variant.

editor/create_asset_bp_camera_sensor_ust.py
```diff
@@ -0,0 +1,253 @@
...
+]
+
+blueprint_desc = \
+{
+    "blueprint_name": "BP_CameraSensor_UST",
+    "blueprint_dir": "/SpContent/Blueprints",
+    "component_descs":
+    [
+        {
+            "name": "final_tone_curve_hdr_",
```
mikeroberts3000, line 147: Match the names to rendering pass names like in the path tracing UST script.

editor/create_asset_bp_camera_sensor_ust.py
```diff
@@ -0,0 +1,253 @@
...
+    # Build the UserSceneTexture universe shared by every capture component. We read each buffer's InternalName
+    # from its own material instance override so it is guaranteed to match what the material actually writes; a
+    # mismatch would assert on the render thread and take the whole engine down.
+    user_scene_texture_material_descs = {}
+    for user_scene_texture_material_name in user_scene_texture_material_names:
+        material_instance_path = posixpath.join("/SpContent/Materials", "MI_PPM_" + user_scene_texture_material_name + "_UST")
+        material_instance = unreal.load_asset(name=material_instance_path)
+        assert isinstance(material_instance, unreal.MaterialInstanceConstant)
+
+        user_scene_texture_overrides = material_instance.get_editor_property(name="user_scene_texture_overrides")
```
mikeroberts3000, line 273: This is a clean approach. Good.

editor/create_asset_bp_camera_sensor_path_tracer_ust.py
```diff
@@ -0,0 +1,213 @@
...
+            "width": width,
+            "height": height,
+            "fov_angle": fov_angle,
+            "num_channels_per_pixel": 4,
+            "channel_data_type": unreal.SpArrayDataType.FLOAT32,
+            "capture_source": unreal.SceneCaptureSource.SCS_FINAL_COLOR_HDR,
+            "show_flag_settings": engine_show_flag_settings["lighting_only_with_path_tracing"],
+            "use_scene_view_extension": True,
+            "override_texture_render_target_format": True,
+            "texture_render_target_format": unreal.TextureRenderTargetFormat.RTF_RGBA32F
```
mikeroberts3000: I don't think we need `RTF_RGBA32F` here. In typical use cases, the only useful data from this component will be from the USTs, which UE forces to be a `float16` data type, so I think we can match the `final_tone_curve_hdr_` render format info here.

editor/create_asset_ppm_ust_instances.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+if __name__ == "__main__":
+
+    # Explicitly load "/SpContent" into the asset registry, since it won't be loaded by default if we are
+    # running as a commandlet, i.e., when the editor is invoked from the command-line with -run=pythonscript
+    # as opposed to -ExecutePythonScript.
+    asset_registry.scan_paths_synchronous(paths=["/SpContent"])
+
+    # For each PPM_* post-process material, create a lightweight material instance that redirects the
```
mikeroberts3000: Correcting inconsistent line width formatting:

```
    # For each PPM_* post-process material, create a lightweight material instance that redirects the material's
    # output to a UserSceneTexture, so a single SpSceneCaptureComponent2D can read the buffer back. We use a
    # material instance rather than a full material copy because the override is resolved on the render proxy at
    # render time, so the instance reuses the parent material's compiled shader with no recompile, and any
    # future edit to the parent material is inherited automatically.
```

editor/create_asset_ppm_ust_instances.py
```diff
@@ -0,0 +1,86 @@
...
+
+        source_material_name = posixpath.split(asset_path)[1].split(".")[0]
+        if not source_material_name.startswith(source_material_name_prefix):
+            continue
+
+        source_material_path = posixpath.join(material_dir, source_material_name)
+        source_material = unreal.load_asset(name=source_material_path)
+        assert isinstance(source_material, unreal.Material)
+
+        # The UserSceneTexture output name is the source material name without the "PPM_" prefix (e.g.,
```
mikeroberts3000, line 43: To avoid confusion, I want the "internal name" specified in the material to be different from the public name in specified in the capture component. So for example, I want the internal name to be `DiffuseColor_UST` and then the public name to be `DiffuseColor`.

editor/create_asset_bp_camera_sensor_path_tracer_ust.py
jakubtomsu: But wait, isn't depth the only valid non-pathtracer-specific UST? I was under the impression that the other channels (e.g. material properties) aren't generated by the pathtracer. But the depth is, since it's required for compositing. It's just written to a regular depth texture instead of a special pathtracer-only one. 

This maps to what the path tracer shader outputs directly:
```hlsl
RWTexture2D<float4> RadianceTexture;
#if PATH_TRACER_USE_ADAPTIVE_SAMPLING
RWTexture2D<float4> VarianceTexture;
#endif
RWTexture2D<float4> AlbedoTexture;
RWTexture2D<float4> NormalTexture;
RWTexture2D<float>  DepthTexture;
```
(from PathTracingCore.ush)

jakubtomsu:
@mikeroberts3000 I updated the example based on your comments. There are also new diffuse only and specular only components for the path tracer. I changed the naming conventions and tried to match the components to the pass names as you mentioned, though I'm not 100% sure everything is as you wanted.

One thing I am unsure about is the example. Currently it works and isn't too complicated, but I'm not sure I like how it combines lumen with the path tracer. It just doesn't feel super clean. But at the same time managing two separate examples might be a bit redundant - the only difference is a couple of path-tracer-specific cvars and a few extra steps in the accumulation loop...

jakubtomsu:
I switched this from a draft PR to a regular one, because it's now in a fully functioning state and mostly includes all the important features. But there's still likely a lot of subtle things we need to improve.

mikeroberts3000:
> _Generator for PPM instances. Fully duplicating the materials would be wasteful, especially since the instances have a way to override the UST name field. Nothing else needs to change on the materials (the resolution divisor can stay at 0!) so this lets us keep the original PPM as a source of truth and re-generate whatever we need. That said I am not sure whether these instances should be checked into source control or generated dynamically._

For better or for worse, our current approach is to check in all of the code-generated assets in SPEAR. I'm in favor of using this approach for the PPM instances. Alternatively, we could avoid checking in these assets, and force users to generate them as one of our setup steps. But I don't mind the current approach. It is convenient for users, and the assets don't change very often so the risk of them becoming stale is small.

> _One thing I am unsure about is the example. Currently it works and isn't too complicated, but I'm not sure I like how it combines lumen with the path tracer. It just doesn't feel super clean. But at the same time managing two separate examples might be a bit redundant - the only difference is a couple of path-tracer-specific cvars and a few extra steps in the accumulation loop..._

It is fine with me if only one or two examples demonstrate this new functionality (e.g., `render_image_hypersim` and maybe a new path tracer example). It is even a good thing to have some examples that use the old type of camera, and some that use this new type of camera, since both approaches may be useful for different users. And the old type of camera uses a somewhat simpler code path, so it is useful for debugging.

I'll have a closer look at how to structure the examples as I'm reviewing this PR.

editor/create_asset_bp_camera_sensor_path_tracer_ust.py
mikeroberts3000: You're right. `SceneDepth` isn't listed in the Path Tracing Buffer Texture material node, so I didn't think it was a path tracer output. But it does appear to be output directly by the path tracer, and therefore it deserves to be included here.

cpp/unreal_plugins/SpUnrealTypes/Source/SpUnrealTypes/SpSceneCaptureComponent2D.h
mikeroberts3000, line 1: Inspired by this PR, I checked this change in already (in this `.h` file), and also fixed up the headers and forward declares in the cpp. So this change is not needed any more.

editor/create_asset_ppm_ust_instances.py
mikeroberts3000, line 1: Rename to `create_asset_mi_ppm_ust_instances.py`.

mikeroberts3000:
This looks good. I slightly modified the code in the `create_asset_bp_camera_sensor_*.py` scripts in your branch.

- Let's use the UST camera implementations in `render_image_hypersim` and `render_image_path_tracer` so we don't need to maintain separate examples demonstrating how to use USTs.
- For the `hypersim` example, match the existing functionality exactly. You won't be able to save `sp_unlit_*` and `sp_object_ids_float16` images because I didn't include those in the UST camera implementation. But otherwise we should be able to match the existing implementation pixel-perfectly.
- For the `path_tracer` example, let's save the primary buffer as well as every UST buffer that is available for the `final_tone_curve_hdr` component. For simplicity, it is fine if we don't output every possible buffer from every possible component.
- There should be an optional `--filter-width` argument (default value of 3.0) that controls `r.PathTracing.FilterWidth`. This cvar affects how primary rays are jittered, and can be used to effectively turn off anti-aliasing for primary rays. This is an important capability to demonstrate, because some intermediate buffers will only composite together correctly when anti-aliasing is turned off for primary rays.
- I'm not sure if it will be relevant for this PR, but I created a `bReadPrimaryPixelData` flag on the capture component (default to `true`) that can be used to toggle whether or not the primary texture target data will be returned to Python. If you find that you are only UST buffer data, without needing the primary data at all, then set `bReadPrimaryPixelData` to false and you won't see (or pay for) a `"data"` buffer in the output returned to Python.

jakubtomsu:
> For the hypersim example, match the existing functionality exactly. You won't be able to save sp_unlit_* and sp_object_ids_float16 images because I didn't include those in the UST camera implementation. But otherwise we should be able to match the existing implementation pixel-perfectly.

Done, but it was quite a substantial change. I did try to set up the new capture config very carefully to preserve the channels in the exact same way.

The path tracer update was a lot simpler, the code is pretty nice and clean IMO. Filter width was also easy, and in the process I read the shader code and learned a lot about importance sampling and why gaussian makes more sense than uniformly jittering the rays within the pixel!

Also it turns out the `bReadPrimaryPixelData` was indeed useful, there's a few passes in hypersim that don't need the main buffer.

jakubtomsu:
Old output:
image

New output:
image

editor/create_asset_ppm_ust_instances.py
mikeroberts3000: Old file still appears in PR.

mikeroberts3000:
Looks great. I fixed a couple of issues in the `create_asset_*` scripts. Be sure to re-run all of these scripts so you're checking in up-to-date assets, and remove `create_asset_ppm_ust_instances.py` because it is a duplicate of `create_asset_mi_ppm_ust_instances.py`.

mikeroberts3000 (approved):
Looks great.

jakubtomsu:
yep all the assets should be updated now

PR #328, opened 2026-07-24
Log output with colors

jakubtomsu:
This is a minor quality of life improvement to the python spear logging, I had this lying around for a while, so I thought why not submit it:)

The reason why I even initially added this was because in some cases, the logs are gigantic and I had trouble visually scanning the terminal output for issues. IMO this makes quite a difference:

image

image

### DIFF
--- a/python/spear/utils/log_utils.py
+++ b/python/spear/utils/log_utils.py
@@ -5,9 +5,23 @@
 
 import inspect
 import os
+import sys
+
+_RESET = "\033[0m"
+_COLORS = {
+    "black":   "30",
+    "red":     "31",
+    "green":   "32",
+    "yellow":  "33",
+    "blue":    "34",
+    "magenta": "35",
+    "cyan":    "36",
+    "white":   "37",
+}
 
 _default_log_enabled = True
 _log_funcs = []
+_supports_color_cached = None
 
 def get_default_log_enabled():
     return _default_log_enabled
@@ -24,11 +38,42 @@ def register_log_func(func):
 def unregister_log_func(func):
     _log_funcs.remove(func)
 
+# supports_color() follows the NO_COLOR/FORCE_COLOR conventions (https://no-color.org/) on top of the standard
+# library's own notion of whether stdout is a terminal (to avoid dumping escape codes to files for example)
+def supports_color():
+    global _supports_color_cached
+    if _supports_color_cached is not None:
+        return _supports_color_cached
+    _supports_color_cached = _supports_color()
+    return _supports_color_cached
+
+def _supports_color():
+    if os.environ.get("NO_COLOR") is not None:
+        return False
+    if os.environ.get("FORCE_COLOR") is not None:
+        return True
+    return sys.stdout.isatty()
+
+def colorize(text, color, bold=False):
+    if not supports_color():
+        return text
+    prefix = "1;" if bold else ""
+    return f"\033[{prefix}{_COLORS[color]}m{text}{_RESET}"
+
+def colorize_message_category(message):
+    message_lower = message.lower()
+    if "error" in message_lower or "exception" in message_lower:
+        print(colorize(text=message, color="red", bold=True))
+    elif "warning" in message_lower:
+        print(colorize(text=message, color="yellow"))
+    else:
+        print(message)
+
 def log(*args):
     current_frame = inspect.currentframe()
     message = _log_get_prefix(current_frame) + "".join([str(arg) for arg in args])
     if _default_log_enabled:
-        print(message)
+        print(colorize_message_category(message))
     for func in _log_funcs:
         func(message)
 
--- a/python/spear/utils/log_utils.py
+++ b/python/spear/utils/log_utils.py
@@ -62,11 +62,11 @@ def colorize(text, color, bold=False):
 def colorize_message_category(message):
     message_lower = message.lower()
     if "error" in message_lower or "exception" in message_lower:
-        print(colorize(text=message, color="red", bold=True))
+        return colorize(text=message, color="red", bold=True)
     elif "warning" in message_lower:
-        print(colorize(text=message, color="yellow"))
+        return colorize(text=message, color="yellow")
     else:
-        print(message)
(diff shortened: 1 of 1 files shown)


python/spear/utils/log_utils.py
```diff
@@ -24,11 +38,41 @@ def register_log_func(func):
...
+    if os.environ.get("NO_COLOR") is not None:
+        return False
+    if os.environ.get("FORCE_COLOR") is not None:
+        return True
+    return sys.stdout.isatty()
+
+def colorize(text, color, bold=False):
+    if not supports_color():
+        return text
+    prefix = "1;" if bold else ""
```
mikeroberts3000, line 59: Don't use ternary ops.

python/spear/utils/log_utils.py
```diff
@@ -24,11 +38,41 @@ def register_log_func(func):
...
+# supports_color() follows the NO_COLOR/FORCE_COLOR conventions (https://no-color.org/) on top of the standard
+# library's own notion of whether stdout is a terminal (to avoid dumping escape codes to files for example)
+def supports_color():
+    global _supports_color_cached
+    if _supports_color_cached is None:
+        _supports_color_cached = _supports_color()
+    return _supports_color_cached
+
+def _supports_color():
+    if os.environ.get("NO_COLOR") is not None:
```
mikeroberts3000, line 50: Use if, elif, else.

python/spear/utils/log_utils.py
```diff
@@ -5,9 +5,23 @@
...
+    "yellow":  "33",
+    "blue":    "34",
+    "magenta": "35",
+    "cyan":    "36",
+    "white":   "37",
+}
 
 _default_log_enabled = True
 _log_funcs = []
+_supports_color_cached = None
```
mikeroberts3000, line 24: `_supports_color_cached` -> `_env_supports_formatting_cached`

python/spear/utils/log_utils.py
```diff
@@ -24,11 +38,41 @@ def register_log_func(func):
 def unregister_log_func(func):
     _log_funcs.remove(func)
 
+# supports_color() follows the NO_COLOR/FORCE_COLOR conventions (https://no-color.org/) on top of the standard
+# library's own notion of whether stdout is a terminal (to avoid dumping escape codes to files for example)
+def supports_color():
```
mikeroberts3000, line 43: These functions are arranged in low-level-to-high-level order. They should be arranged in the opposite order.

mikeroberts3000, line 43: `supports_color` -> `_env_supports_formatting`

python/spear/utils/log_utils.py
```diff
@@ -24,11 +38,41 @@ def register_log_func(func):
...
+    global _supports_color_cached
+    if _supports_color_cached is None:
+        _supports_color_cached = _supports_color()
+    return _supports_color_cached
+
+def _supports_color():
```
mikeroberts3000, line 49: `_supports_color` -> `_env_supports_formatting_impl`

python/spear/utils/log_utils.py
```diff
@@ -24,11 +38,41 @@ def register_log_func(func):
...
+        return True
+    return sys.stdout.isatty()
+
+def colorize(text, color, bold=False):
+    if not supports_color():
+        return text
+    prefix = "1;" if bold else ""
+    return f"\033[{prefix}{_COLORS[color]}m{text}{_RESET}"
+
+def colorize_message_category(message):
```
mikeroberts3000, line 62: `colorize_message_category` -> `_request_format`

python/spear/utils/log_utils.py
```diff
@@ -24,11 +38,41 @@ def register_log_func(func):
...
+    return _supports_color_cached
+
+def _supports_color():
+    if os.environ.get("NO_COLOR") is not None:
+        return False
+    if os.environ.get("FORCE_COLOR") is not None:
+        return True
+    return sys.stdout.isatty()
+
+def colorize(text, color, bold=False):
```
mikeroberts3000, line 56: `colorize` -> `_format`
`text` -> `message`

python/spear/utils/log_utils.py
```diff
@@ -5,9 +5,23 @@
 
 import inspect
 import os
+import sys
+
+_RESET = "\033[0m"
+_COLORS = {
```
mikeroberts3000, line 11: Make sure that the colors you're using make sense semantically (red should be more urgent than orange), and that they are readable on black and white backgrounds (the default terminal on macOS has a white background).

python/spear/utils/log_utils.py
```diff
@@ -24,11 +38,41 @@ def register_log_func(func):
...
+        return colorize(text=message, color="yellow")
+    else:
+        return message
+
 def log(*args):
     current_frame = inspect.currentframe()
     message = _log_get_prefix(current_frame) + "".join([str(arg) for arg in args])
     if _default_log_enabled:
-        print(message)
+        print(colorize_message_category(message))
```
mikeroberts3000, line 75: Our convention is to always use kwargs when calling functions that are implemented in this codebase.

```
print(_request_format(message=message))
```

mikeroberts3000:
After thinking about this PR in more detail, I've decided to decline it. Most of our Python error reporting happens through `assert` statements; not exceptions, or log messages with the word `ERROR` in them, etc. For example, `assert` occurs 468 times in the `python` folder, versus `ERROR` which occurs 29 times. So this PR is useful, but it it also adds complexity, and will only highlight less than 10% of the possible Python errors that might occur in the code. Additionally, it doesn't do anything to highlight error or assert messages in the C++ code, meaning that 10% is a significant over-estimate.

Users that want custom logging like this can still achieve it in their own code by registering their own logging function using `spear.register_log_func(...)` and disabling the default one using `spear.set_default_log_enabled(...)`, so I don't think the functionality in this PR needs to be in core library code.

jakubtomsu:
Okay, this is fair! I understand the decision

PR #333, opened 2026-08-24
PT glass material

jakubtomsu:
Adds a very simple PT-compatible glass base shader to `SpContent`. Unlike most glass materials made for lumen that only model alpha-based translucency, this material does actual light refraction. This material can either be instantiated directly or used as a reference point for custom materials. I made this based on photo references and physically based values, so it should be relatively close to actual real world glass.

image
image


mikeroberts3000:
This looks good. I don't think it's fundamental enough to have in `SpContent` though (e.g., we don't refer to it anywhere in the plugin code). Let's move it and rename it to `/Game/SPEAR/Materials/M_Glass_PT`.

mikeroberts3000 (approved):
Looks good. Please apply my suggested changes before merging.

mikeroberts3000:
Rename to `M_Glass_PT`.

PR #335, opened 2026-08-24
PPM opacity output

jakubtomsu:
The PPMs already have the alpha output flag enabled, but it's not properly hooked up to read the actual GBuffer alpha channel data. Opacity output is required for things like holdout rendering and masking out render layers.

Note: the denoiser does not preserve nor denoise the alpha channel, IIRC it just sets it to pure white. For this reason the `PPM_PathTracingDenoisedRadiance` reads from regular (non-denoised) radiance texture. This is the best approach in practice because the alpha channel would otherwise go unused, so this way we don't need to do a separate render (non-UST case) or read a full extra texture (UST case).

Since git cannot display uasset diff, here is what the new PPMs look like. Previously, the alpha output was not connected to anything and there was only the RGB output.

image
image


mikeroberts3000:
- Should we do this for other `PPMs`? Is holdout rendering etc something that someone might want to do using the rasterizer? Is that even possible?
- If we change this, do we need to re-run the code to generate the UST variants? Or do the UST variants inherit this change implicitly.

jakubtomsu:
> Should we do this for other PPMs? Is holdout rendering etc something that someone might want to do using the rasterizer? Is that even possible?

Alpha holdout should be available on Lumen, I believe it was initially PT-only but Lumen support was added in 5.5.

> If we change this, do we need to re-run the code to generate the UST variants? Or do the UST variants inherit this change implicitly.

I believe all the `MI_PPM_*` don't need to be regenerated, the engine knows about their direct dependency and should just update the cached resources as necessary. I think material instances should share the PSO anyway and only set the shader uniform variables, CPU side properties and stuff like that, so it should just work...

mikeroberts3000:
Let's hook up alpha correctly for all of the `PPM` materials. I believe the type of SceneTexture.Color is a float4, so for all of the path-tracer materials and for all of the regular scene texture materials, it should be possible to break up Color into RGB and A and hook A up to the opacity. Doing so might be useful for someone, and it's better to output all possible data and let the user sort through it later, since they're paying for all 4 channels of memory bandwidth consumption anyway. And it's better to apply this change consistently to avoid confusion.

jakubtomsu:
okay, just went through all the materials and hooked up the opacity. In some cases, it's clearly a single channel output, so there I broadcasted the value to alpha channel as well, since that's what already happens for the RGB channels. There were a few PPMs I didn't change at all, for example world position etc, because in that case there is nothing to "broadcast" and the alpha channel just doesn't mean anything.

Many of the textures probably don't have an actual alpha component but yeah that doesn't really matter, we just want to accurately transfer the data.

mikeroberts3000:
Agreed with your approach in all cases. If the data is coming from a scene texture, then its static type is a float4 and should map to RGBA. But if the data has a static type of float3, then it only makes sense to map to RGB and there is nothing to map to A. So the consistent pattern is that we always use the static type to decide how to map. In cases where a g-buffer is "clearly" a scalar (e.g., metallic), that is a fact about its dynamic type. The static type of the output pin of the scene texture node is still a float4, so we still map to RGBA.

mikeroberts3000 (approved):
Great.

Issue #332, opened 2026-08-20
ue-58: build_third_party_libs.py fails on a binary UE 5.8 install (libc++ moved out of the engine tree; libc++abi not linked)


Color2333:
### Summary

On the `ue-58` branch, `tools/build_third_party_libs.py` fails on Linux against a **binary** Unreal Engine 5.8 install. Two independent causes, both consequences of UE 5.8 moving libc++ out of the engine source tree. Small fix at the end; verified locally.

### Environment

| | |
|---|---|
| Branch | `ue-58` |
| Engine | UE **5.8.1**, official Linux binary (`Linux_Unreal_Engine_5.8.1.zip`), not built from source |
| Toolchain | engine-shipped `v26_clang-20.1.8-rockylinux8` |
| Host | Rocky-compatible x86_64, Python 3.12 |

### Problem 1 — boost builds nothing: the libc++ include path no longer exists

`build_third_party_libs.py` derives

```
Engine/Source/ThirdParty/Unix/LibCxx/include/c++/v1
Engine/Source/ThirdParty/Unix/LibCxx/lib/Unix/x86_64-unknown-linux-gnu
```

In UE 5.8 **`Engine/Source/ThirdParty/Unix/` does not exist at all** (only `Engine/Source/ThirdParty/Licenses/libcxx_v18.1.0.license` remains). libc++ now ships inside the clang SDK the script already locates:

```
Engine/Extras/ThirdPartyNotUE/SDKs/HostLinux/Linux_x64/v26_clang-20.1.8-rockylinux8/
    x86_64-unknown-linux-gnu/include/c++/v1     # 188 headers
    x86_64-unknown-linux-gnu/lib64/libc++.a
    x86_64-unknown-linux-gnu/lib64/libc++abi.a
```

Because the flags are `-nostdinc++ -I<nonexistent>`, the standard library is disabled and nothing replaces it, so every `#include` fails. The failure is **silent about the real cause** — `b2` just reports

```
...failed updating 0 target...
...skipped <p...>libboost_unit_test_framework.a for lack of ...
```

followed by `CalledProcessError` from `subprocess.run(..., check=True)`. Nothing says "that include directory is empty/missing".

This path was introduced by #294 (*use Unreal shipped clang and libc++ on Linux*), which was correct for the 5.5 layout.

### Problem 2 — yaml-cpp fails to link: `libc++abi` is not on the link line

[...]

PR #336, opened 2026-08-25
ue-58: fix third-party build against a binary UE 5.8 install (libc++ SDK paths, -lc++abi)

Color2333:
Fixes #332.

Two changes to `tools/build_third_party_libs.py` so that the `ue-58` branch builds against a **binary** UE 5.8 install on Linux. Both are consequences of UE 5.8 moving libc++ out of the engine source tree.

### 1. libc++ paths now point at the clang SDK

`Engine/Source/ThirdParty/Unix/` no longer exists in UE 5.8; libc++ ships inside the clang SDK the script already locates. Per your comment I dropped the legacy fallback entirely and made `ue-58` support the new layout only.

The asserts check that the directories are **non-empty**, not merely that the paths exist. This matters because the compile flags are `-nostdinc++ -I<include_dir>`: if that directory is missing *or empty*, the standard library is disabled with nothing replacing it, every `#include` fails, and `b2` reports only

```
...failed updating 0 target...
```

which never mentions the include path. An assert at the point of derivation turns a silent 5-minute failure into an immediate message.

### 2. `-lc++abi` on the yaml-cpp link line, and `YAML_CPP_BUILD_TOOLS=OFF`

The SDK ships libc++ and libc++abi as two separate static libraries in the same `lib64`, and in this configuration (`-nostdinc++` plus a custom `-L`) `-stdlib=libc++` only pulls in `-lc++`.

### On `CMAKE_EXE_LINKER_FLAGS` — you were right to be suspicious

You asked why we set it at all once no executables are built, and whether something like `CMAKE_STATIC_LIB_LINKER_FLAGS` would be more correct. I tested rather than guessed. Configuring and building yaml-cpp with `YAML_CPP_BUILD_TESTS=OFF YAML_CPP_BUILD_TOOLS=OFF`, once with the flags and once with the variable omitted entirely:

* Both configure and build cleanly.
* **The resulting `libyaml-cpp.a` is byte-identical** (`cmp` reports no difference; 899 defined symbols either way).

So it has no effect on the artifact. Its one remaining effect is on CMake's configure-time compiler-ABI probe, which *does* link a small executable:

```
with the flags:     CMAKE_CXX_IMPLICIT_LINK_LIBRARIES = c++abi;pthread;c++;m;gcc_s;gcc;c;...
without the flags:  CMAKE_CXX_IMPLICIT_LINK_LIBRARIES = stdc++;m;gcc_s;gcc;c;...
```

Without them CMake records **libstdc++** as the implicit C++ runtime for a build that is compiled against libc++ headers. That value is only consumed by CMake-driven link steps, and SPEAR consumes the `.a` through UBT, so nothing acts on it today — but it is wrong, and it would become load-bearing again the moment any executable target is re-enabled.

On the variable name: `CMAKE_STATIC_LIB_LINKER_FLAGS` does not exist. The real one is [`CMAKE_STATIC_LINKER_FLAGS`](https://cmake.org/cmake/help/latest/variable/CMAKE_STATIC_LINKER_FLAGS.html), documented as *"passed to the archiver when creating a static library"* — and the archive step here is

```
llvm-ar qc libyaml-cpp.a <objects...>
```

No linker is involved, so `-lc++abi` there would be handed to `ar` and treated as a filename. There is no better-named option for this, because there is nothing to link.

I kept `CMAKE_EXE_LINKER_FLAGS` with `-lc++abi` added, on the grounds that it keeps CMake's own probe consistent with the toolchain and keeps the build correct if tools or tests are ever switched back on. **If you would rather see it deleted, say the word — the evidence above says the artifact will not change, and I will push that instead.**

### Verified

```console
python tools/build_third_party_libs.py --unreal-engine-dir /path/to/UE_5.8.1
```

on UE **5.8.1** official Linux binary, engine-shipped `v26_clang-20.1.8-rockylinux8`, Rocky-compatible x86_64, CMake 3.28.3, produces:

```
third_party/boost/stage/lib/libboost_filesystem.a
third_party/boost/stage/lib/libboost_unit_test_framework.a
third_party/rpclib/BUILD/Linux/librpc.a
third_party/yaml-cpp/BUILD/Linux/libyaml-cpp.a
```

Not tested on Windows or macOS; both code paths are untouched. Not tested against a source build of UE 5.8 — if a source build still populates `Engine/Source/ThirdParty/Unix/LibCxx`, this branch now ignores it in favour of the SDK copy, which I believe is what you intended.

---
🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/tools/build_third_party_libs.py
+++ b/tools/build_third_party_libs.py
@@ -104,9 +104,20 @@ if __name__ == "__main__":
 
         spear.log("Found Unreal clang: ", linux_clang_path)
 
+        # In UE 5.8, libc++ ships inside the clang SDK we just located, rather than in
+        # Engine/Source/ThirdParty/Unix/LibCxx as it did in UE 5.5. Note that the compiler
+        # flags below are "-nostdinc++ -I<linux_libcpp_include_dir>", so if the include
+        # directory is wrong, the standard library is disabled and nothing replaces it,
+        # and every #include fails. We assert that the directories are non-empty rather
+        # than merely present, because an empty directory reproduces the same failure.
         linux_clang_bin_dir      = os.path.realpath(os.path.join(linux_clang_path, "x86_64-unknown-linux-gnu", "bin"))
-        linux_libcpp_include_dir = os.path.realpath(os.path.join(unreal_engine_dir, "Engine", "Source", "ThirdParty", "Unix", "LibCxx", "include", "c++", "v1"))
-        linux_libcpp_lib_dir     = os.path.realpath(os.path.join(unreal_engine_dir, "Engine", "Source", "ThirdParty", "Unix", "LibCxx", "lib", "Unix", "x86_64-unknown-linux-gnu"))
+        linux_libcpp_include_dir = os.path.realpath(os.path.join(linux_clang_path, "x86_64-unknown-linux-gnu", "include", "c++", "v1"))
+        linux_libcpp_lib_dir     = os.path.realpath(os.path.join(linux_clang_path, "x86_64-unknown-linux-gnu", "lib64"))
+
+        assert os.path.isdir(linux_libcpp_include_dir) and os.listdir(linux_libcpp_include_dir), f"Could not find libc++ headers: {linux_libcpp_include_dir}"
+        assert glob.glob(os.path.join(linux_libcpp_lib_dir, "libc++.*")), f"Could not find libc++ library: {linux_libcpp_lib_dir}"
+
+        spear.log("Found Unreal libc++: ", linux_libcpp_include_dir)
 
         boost_toolset = "clang"
         boost_toolset_version = ""
@@ -343,15 +354,20 @@ if __name__ == "__main__":
 
     elif sys.platform == "linux":
 
-        # -lpthread needed on some Linux environments
+        # The clang SDK in UE 5.8 ships libc++ and libc++abi as two separate static
+        # libraries, and in this configuration "-stdlib=libc++" only pulls in "-lc++", so
+        # we need "-lc++abi" to resolve the ABI symbols. -lpthread needed on some Linux
+        # environments. We also set YAML_CPP_BUILD_TOOLS=OFF, because we only need the
+        # yaml-cpp library, not its command-line tools.
         cmd = \
             "cmake " + \
             f'"-DCMAKE_CXX_COMPILER={cxx_compiler}" ' + \
             f'"-DCMAKE_POSITION_INDEPENDENT_CODE=ON" ' + \
             f'"-DCMAKE_CXX_FLAGS={cmake_cxx_flags}" ' + \
-            f'"-DCMAKE_EXE_LINKER_FLAGS=-stdlib=libc++ -L\'{linux_libcpp_lib_dir}\' -lpthread" ' + \
+            f'"-DCMAKE_EXE_LINKER_FLAGS=-stdlib=libc++ -L\'{linux_libcpp_lib_dir}\' -lc++abi -lpthread" ' + \
             f'"-DCMAKE_VERBOSE_MAKEFILE={cmake_verbose_makefile}" ' + \
             '"-DYAML_CPP_BUILD_TESTS=OFF" ' + \
+            '"-DYAML_CPP_BUILD_TOOLS=OFF" ' + \
             '"-DCMAKE_POLICY_VERSION_MINIMUM=3.5" ' + \
             os.path.join("..", "..")
 
--- a/tools/build_third_party_libs.py
+++ b/tools/build_third_party_libs.py
@@ -104,18 +104,12 @@ if __name__ == "__main__":
 
         spear.log("Found Unreal clang: ", linux_clang_path)
 
-        # In UE 5.8, libc++ ships inside the clang SDK we just located, rather than in
-        # Engine/Source/ThirdParty/Unix/LibCxx as it did in UE 5.5. Note that the compiler
-        # flags below are "-nostdinc++ -I<linux_libcpp_include_dir>", so if the include
-        # directory is wrong, the standard library is disabled and nothing replaces it,
-        # and every #include fails. We assert that the directories are non-empty rather
-        # than merely present, because an empty directory reproduces the same failure.
         linux_clang_bin_dir      = os.path.realpath(os.path.join(linux_clang_path, "x86_64-unknown-linux-gnu", "bin"))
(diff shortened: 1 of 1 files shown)


tools/build_third_party_libs.py
```diff
@@ -104,9 +104,20 @@
 
         spear.log("Found Unreal clang: ", linux_clang_path)
 
+        # In UE 5.8, libc++ ships inside the clang SDK we just located, rather than in
```
mikeroberts3000: Comment not necessary.

tools/build_third_party_libs.py
```diff
@@ -104,9 +104,20 @@
...
+        # directory is wrong, the standard library is disabled and nothing replaces it,
+        # and every #include fails. We assert that the directories are non-empty rather
+        # than merely present, because an empty directory reproduces the same failure.
         linux_clang_bin_dir      = os.path.realpath(os.path.join(linux_clang_path, "x86_64-unknown-linux-gnu", "bin"))
-        linux_libcpp_include_dir = os.path.realpath(os.path.join(unreal_engine_dir, "Engine", "Source", "ThirdParty", "Unix", "LibCxx", "include", "c++", "v1"))
-        linux_libcpp_lib_dir     = os.path.realpath(os.path.join(unreal_engine_dir, "Engine", "Source", "ThirdParty", "Unix", "LibCxx", "lib", "Unix", "x86_64-unknown-linux-gnu"))
+        linux_libcpp_include_dir = os.path.realpath(os.path.join(linux_clang_path, "x86_64-unknown-linux-gnu", "include", "c++", "v1"))
+        linux_libcpp_lib_dir     = os.path.realpath(os.path.join(linux_clang_path, "x86_64-unknown-linux-gnu", "lib64"))
+
+        assert os.path.isdir(linux_libcpp_include_dir) and os.listdir(linux_libcpp_include_dir), f"Could not find libc++ headers: {linux_libcpp_include_dir}"
```
mikeroberts3000: The asserts are good. Our convention in this codebase is to not use assert messages though because they're a maintenance hazard.

tools/build_third_party_libs.py
```diff
@@ -343,15 +354,20 @@
 
     elif sys.platform == "linux":
 
-        # -lpthread needed on some Linux environments
+        # The clang SDK in UE 5.8 ships libc++ and libc++abi as two separate static
+        # libraries, and in this configuration "-stdlib=libc++" only pulls in "-lc++", so
+        # we need "-lc++abi" to resolve the ABI symbols. -lpthread needed on some Linux
+        # environments. We also set YAML_CPP_BUILD_TOOLS=OFF, because we only need the
+        # yaml-cpp library, not its command-line tools.
```
mikeroberts3000: Comment not necessary. I think we should use `CMAKE_STATIC_LIB_LINKER_FLAGS` here instead of `CMAKE_EXE_LINKER_FLAGS` because it is more semantically meaningful with respect to what we're trying to do. Also, we should apply `YAML_CPP_BUILD_TOOLS=OFF` on all platforms. I will test on Windows and macOS. It is fine with me if the rest of this change only touches the Linux path.

mikeroberts3000:
Hi. I suggested some minor changes but otherwise this looks good! I'm happy to merge once those changes are addressed. Thank you ☺️

tools/build_third_party_libs.py
Color2333: Removed.

tools/build_third_party_libs.py
Color2333: Done — assert messages dropped.

tools/build_third_party_libs.py
Color2333: Comment removed, and `YAML_CPP_BUILD_TOOLS=OFF` is now applied on all three platforms.

On `CMAKE_STATIC_LIB_LINKER_FLAGS` — I tried it, and I don't think it can work. Two problems:

**That variable does not exist.** CMake accepts it silently and then reports it as unused:

```
CMake Warning:
  Manually-specified variables were not used by the project:

    CMAKE_STATIC_LIB_LINKER_FLAGS
```

The real one is [`CMAKE_STATIC_LINKER_FLAGS`](https://cmake.org/cmake/help/latest/variable/CMAKE_STATIC_LINKER_FLAGS.html).

**And the real one breaks the build,** because it goes to the archiver, not a linker:

```console
$ cmake -DCMAKE_STATIC_LINKER_FLAGS=-lc++abi ... && cmake --build .
llvm-ar qc libyaml-cpp.a -lc++abi "CMakeFiles/yaml-cpp.dir/src/binary.cpp.o" ...
llvm-ar: error: unknown option +
gmake[2]: *** [libyaml-cpp.a] Error 1
```

`ar` has no notion of `-l`, so it reads `-lc++abi` as a flag string and fails.

**So I removed the linker flags instead.** I think that is what your "more semantically meaningful" instinct was actually pointing at: with `YAML_CPP_BUILD_TOOLS=OFF` and `YAML_CPP_BUILD_TESTS=OFF` there is no link step at all, so there is nothing for `-lc++abi` (or `-lpthread`) to do. I verified that the artifact is unchanged — configuring with and without `CMAKE_EXE_LINKER_FLAGS` produces a **byte-identical** `libyaml-cpp.a` (`cmp` reports no difference, 899 defined symbols either way).

The one thing that does change is CMake's configure-time ABI probe, which links a small executable of its own:

```
with the flags:     CMAKE_CXX_IMPLICIT_LINK_LIBRARIES = c++abi;pthread;c++;m;gcc_s;gcc;c;...
without the flags:  CMAKE_CXX_IMPLICIT_LINK_LIBRARIES = stdc++;m;gcc_s;gcc;c;...
```

Nothing consumes that value here — SPEAR takes the `.a` through UBT — so I left it alone rather than adding flags back to satisfy a probe. Flagging it in case you would rather keep the flags for that reason; happy to restore them.

One caveat I could not test: the deleted comment said `-lpthread` was "needed on some Linux environments". Since no executable is linked any more that should be moot, but it was presumably added for a real failure on a machine I don't have.

Full build re-verified on UE 5.8.1 after these changes: boost, rpclib and yaml-cpp all build.

mikeroberts3000 (approved):
Looks good. Thank you!

mikeroberts3000:
I think the resolve conflicts view is getting confused. It should be straightforward to merge this PR the new changes in the `ue-58` branch from a few mins ago though. Please do this carefully before merging.

mikeroberts3000:
I was in the `ue-58` branch anyway so I included this change as I was dealing with other conflicts, and I also propagated the `-lc++abi` fix to `install_python_extension.py` since I think it will be necessary there too.

PR #337, opened 2026-08-27
Support all light override showflags

jakubtomsu:
Matches light override support with other similar engine components (e.g. `GameViewportClient`)

The existing old behavior is preserved, but now all the `showflags` users expect to work are supported.


mikeroberts3000 (approved):
Perfect.

PR #339, opened 2026-09-09
Fix glass material

jakubtomsu:
For some reason the material I authored in #333 is broken, it looks like it wasn't working because I renamed it from `M_GlassPT` to `M_Glass_PT` in the file explorer instead of using the UE Content Browser. I guess that caused a validation error causing the material to be invisible in the editor even when the file existed. Sorry for the mistake. It should now be good.


mikeroberts3000 (approved):
Looks good. Unreal is annoying about editing content outside of the editor. Haha.

